PR #5, opened 2015-09-19
Adds a way to automatically add a Spark.function tied to an instance of one of our Equipment classes at construction time.

kyleoliveira:
Note, however, that the function names rely on the ID of the instance and we still haven't implemented a good way to set that in the case of compound classes (i.e. Thermostat and RIMS). For now, you'll need to keep that in mind. For example, if you were to create a RIMS and then a Pump (say, in `setup()`), you'll want to give the Pump an ID of '2' because the Pump associated with Spark.function `pump_1` is part of the RIMS object.


kyleoliveira:
Most recent three commits were in response to a [post](https://community.particle.io/t/declaring-a-spark-function-in-the-constructor-of-a-subclass/15749) I put up on the Particle community boards asking for a good way to do this sort of thing when dealing with derived classes. Much cleaner now.

PR #10, opened 2015-09-21
Adds support for the Dallas OneWire sensor.

azyth:
integrated DS18B20 onewire libraries. 

initial code for single probe connected to rhizome.


kyleoliveira:
Ok, so, for starters...
It looks like you're trying to use the [particle-ds18x20](https://github.com/Ohmbrewer/particle-ds18x20) library. Supposing we do want to use that library, you should include it like we do the Adafruit_ILI9341 library - namely:
1. Pull the git repo down in the same directory as you have your Rhizome repo.
2. Add references to all of the source and header files to the CMakeLists.txt file
3. Finally, reference the appropriate .h files in our code (e.g. at the top of TemperatureSensor.cpp, probably)

The reasoning behind this is two-fold. First, it means that if there's an update to that library, we can just do a git pull on that repo and we don't have to do some sort of complex merge into our Rhizome code. Second, it keeps this Rhizome repo focused on Rhizome stuff.

It's actually a lot easier to do it this way - just look at how I integrated that aforementioned Adafruit library (and the library it relies on, in fact). Not hairy at all.

kyleoliveira:
Further, I'm not sure this is the best library to use. Like it says in the README for the library, this is regular C and all the functions pop into the global namespace. It feels like while that might be easier to get off the ground, it will probably cause problem in the long-run.

It might be more appropriate to use one of the OneWire libraries available via the Build IDE (either [OneWire](https://github.com/Hotaman/OneWireSpark) or [spark-dallas-temperature](https://github.com/tomdeboer/SparkCoreDallasTemperature)) as these seem like they'd be more likely to be maintained and also have the advantage of being written using C++ objects. Again, you'd include them in the same way as I have for the Adafruit libraries.

Give that a shot and if there's no good way to make one of those libraries work, we can circle back around and see if we can do something to protect ourselves against the issues I mentioned.

lib/Ohmbrewer_Temperature_Sensor.cpp
```diff
@@ -124,11 +124,82 @@ bool Ohmbrewer::TemperatureSensor::isOff() const {
...
+     */
+    uint8_t numsensors = ow_search_sensors(10, sensors);
+    sprintf(msg, "Found %i sensors", numsensors);
+    log(msg);
+    if ( DS18X20_read_meas( &sensors[0], &subzero, &cel, &cel_frac_bits) == DS18X20_OK ) {
+        char sign = (subzero) ? '-' : '+';
+        int frac = cel_frac_bits*DS18X20_FRACCONV;
+        temp_c = (double)cel;
+       // temp_c = temp_c + (10**(-4)*(double)frac);  //TODO integrate fraction to end of temp_c
+        Temperature::fromC(temp_c);
```
kyleoliveira: This isn't quite right... You'll want to say something like:

```
getTemp()->fromC(temp_c);
```

If that looks too wrong, you can wrap it with a new function along the lines of `setTempFromC(double celsiusTemp)`.

lib/Ohmbrewer_Temperature_Sensor.cpp
```diff
@@ -124,11 +124,82 @@ bool Ohmbrewer::TemperatureSensor::isOff() const {
...
+                {
+                    // Do Nothing, error?
+                }
+            }
+        }
+    }
+
+
+     */
+    return (int)(millis()-start_time);
```
kyleoliveira: millis() already returns an int, so you shouldn't have to cast this. Just fix the definition above.

azyth:
Regarding Integrating the Library, I had tried to do it that way first, but didnt see how to do it , (the make file is the key) 

This is one of the libraries in the build IDE, I forked it directly from particle. 

I am open to using a different lib, however this one seemed alot easier to use for multi probes then the ones i was trying to use this summer(the others you mention)

lib/Ohmbrewer_Temperature_Sensor.cpp
azyth: gettemp() returns the last temperature read by the sensor not a pointer to the location of the sotred temp....right?

My main problem is I cannot find where the temperature is stored to update with fromC().  I need access to some temperature object on which to call fromC().

lib/Ohmbrewer_Temperature_Sensor.cpp
```diff
@@ -72,6 +79,22 @@ Ohmbrewer::TemperatureSensor::~TemperatureSensor() {
...
+void Ohmbrewer::TemperatureSensor::setProbeId(char id[8]){
+    _probeId = id;
+}
+/**
+ * reads current probes connected to busPin
+ * @param - sensors array to update with ID values of all(max10) the connected ds18b20 probes
+ * @returns - number of sensors discovered
+ */
+int Ohmbrewer::TemperatureSensor::findProbeIds(uint8_t sensors[80]){
+    return (int)ow_search_sensors(10, sensors);
```
kyleoliveira: Why cast this?

kyleoliveira: Also, it looks like this function only returns the number of sensors... Am I wrong?

lib/Ohmbrewer_Temperature_Sensor.cpp
```diff
@@ -149,11 +172,87 @@ bool Ohmbrewer::TemperatureSensor::isOff() const {
...
+     */
+
+    if ( DS18X20_read_meas( &sensors[0], &subzero, &cel, &celFracBits) == DS18X20_OK ) {
+        char sign = (subzero) ? '-' : '+';
+        int frac = celFracBits*DS18X20_FRACCONV;
+        tempC = (double)cel;
+        if (sign == '-'){
+            tempC = tempC * -1;
+        }
+        tempC = tempC + (.0001*(double)frac);
```
kyleoliveira: It would be good to extract this conversion stuff into a a function, if only for readability.

lib/Ohmbrewer_Temperature_Sensor.cpp
```diff
@@ -212,3 +311,11 @@ void Ohmbrewer::TemperatureSensor::whichPins(std::list<int>* pins) {
...
 }
 
+/*
+ * publish to the log
+ */
+void log(char* msg)
+{
+    Spark.publish("log", msg);
+    delay(500);
+}
```
kyleoliveira: Make sure to have a blank line at the end of your code files.

lib/Ohmbrewer_Temperature_Sensor.h
```diff
@@ -178,6 +192,11 @@ namespace Ohmbrewer {
              * Digital Pin for the temperature bus one wire protocol
              */
             int _busPin;
+
+            /**
+             * Unique ID for the temperature probe [8] char array ID code
+             */
+            char* _probeId;
```
kyleoliveira: Why do we need a _busPin and a _probeId if we can only really intend to support one OneWire bus? It seems like it would be simpler to rename _busPin to something like _busIndex and use it to indicate where in the list of sensors the probe for the given TemperatureSensor is. Then we could use `ow_find_sensor` to pull a reading off that one alone. Yae/Nay?

lib/Ohmbrewer_Temperature_Sensor.cpp
```diff
@@ -149,11 +172,87 @@ bool Ohmbrewer::TemperatureSensor::isOff() const {
...
+				sensors[(i*OW_ROMCODE_SIZE)+3],
+				sensors[(i*OW_ROMCODE_SIZE)+4],
+				sensors[(i*OW_ROMCODE_SIZE)+5],
+				sensors[(i*OW_ROMCODE_SIZE)+6],
+				sensors[(i*OW_ROMCODE_SIZE)+7]};
+
+        if (sensors[i*OW_ROMCODE_SIZE+0] == 0x10 || sensors[i*OW_ROMCODE_SIZE+0] == 0x28) //0x10=DS18S20, 0x28=DS18B20
+        {
+            // if current probe matches the probe we are looking for
+            if (_probeId == probeId){
```
kyleoliveira: This isn't going to work for a c-string. You'd need to to strcmp().

lib/Ohmbrewer_Temperature_Sensor.cpp
```diff
@@ -149,11 +172,87 @@ bool Ohmbrewer::TemperatureSensor::isOff() const {
...
+        char probeId[8] = {
+				sensors[(i*OW_ROMCODE_SIZE)+0],
+				sensors[(i*OW_ROMCODE_SIZE)+1],
+				sensors[(i*OW_ROMCODE_SIZE)+2],
+				sensors[(i*OW_ROMCODE_SIZE)+3],
+				sensors[(i*OW_ROMCODE_SIZE)+4],
+				sensors[(i*OW_ROMCODE_SIZE)+5],
+				sensors[(i*OW_ROMCODE_SIZE)+6],
+				sensors[(i*OW_ROMCODE_SIZE)+7]};
+
```
kyleoliveira: If we are recording the whole probe ID, we should only need to pull it once. If you're going to leave these comments in, move them into a separate method comment it out.

lib/Ohmbrewer_Temperature_Sensor.cpp
```diff
@@ -212,3 +311,11 @@ void Ohmbrewer::TemperatureSensor::whichPins(std::list<int>* pins) {
 
 }
 
+/*
+ * publish to the log
+ */
+void log(char* msg)
```
kyleoliveira: If you're going to use this, you need to add the namespace on the function and include the function in the header file. Otherwise you're just inserting a global function, and (despite how this OneWire library works), we're trying to stay fairly object-oriented.

README.md
```diff
@@ -10,3 +10,15 @@ file. Simply add a ```build``` directory to the project directory and navigate t
...
+
+### Sensor_*
+Ported AVR code for interacting with DS18B20 sensors.
+
+Supports multiple sensors on one pin.
+
+Standard hardware configuration, 4.7k ohm resistor from data to VCC line.
+
+Note that this is C code, not CPP, so functions etc are not enclosed in their own name space.
+The functions are named crc8 for the CRC calculations, ow_ for onewire functions, and DS18X20_ for temperature specific functions.
```
kyleoliveira: It looks like you copied most of this new stuff directly from the README for the particle-ds18x20 library. I don't think it's really necessary here - if you need to know how to use the library correctly, you're going to look at the library itself.

lib/Ohmbrewer_Temperature_Sensor.cpp
azyth: i was casting it to int so that the callee could deal with it as a normal number.  

and search sensors returns num sensors but updates sensors with all the IDs

lib/Ohmbrewer_Temperature_Sensor.h
azyth: This is useful gogin forward for when we support multiple sensor types and the sensor may be reading on a different pin for a RTD sensor  which dosnt need an ID but will have a bus pin. ....etc.

lib/Ohmbrewer_Temperature_Sensor.cpp
azyth: ...right

lib/Ohmbrewer_Temperature_Sensor.cpp
azyth: i dont get what you mean, this code block is the functional code for this method. and we have to pull it everytime we read in order to establish that our probe is still connected. 
I am leaving this comented out here so that once we add the set up capibilities elsewhere in the program we can easily change the fnutionality.
Unless you mean make a duplicate function that we can change the call to when we do the above mentioned.?

lib/Ohmbrewer_Temperature_Sensor.cpp
azyth: somehow this function magically disappeared from my code , my file ends at line 288.....weird. 
damn git abnormalities. re pulling`

lib/Ohmbrewer_Temperature_Sensor.cpp
azyth: really? add a whole function to eliminate 2 lines of code?

kyleoliveira: Yeah, and it'd drop this down to 1 line within the if-statement, right?:

```
getTemp()->fromC(convertDS18X20Reading(&sensors[0], &subzero, &cel, &celFracBits))
```

Whatever you call this method, it should be static if you did it this way since it's doesn't concern a specific instance of TemperatureSensor.

lib/Ohmbrewer_Temperature_Sensor.cpp
kyleoliveira: Do we need this every time though? The probe ID is the same every time. Look at `ow_find_sensor`, it seems to provide a way to pull the sensor you want given a probe ID. There are also definitions to compare against to check if that function fails (see `ow_search_sensors` for an example).

What I'm suggesting is that when you add the function for adding a OneWire probe, it should store the probe ID. Then you don't have to cycle through all the devices on the OneWire bus when you're doing a reading for a specific probe.

lib/Ohmbrewer_Temperature.h
```diff
@@ -8,6 +8,7 @@
 
 #include "application.h"
 
+
```
kyleoliveira: If you drop this space this file won't get touched in the commit... just sayin'...

kyleoliveira:
Closing in favor of #24

PR #14, opened 2015-09-23
Adds Temperature::toStrC and Temperature::toStrF for a standardized way to get the C-string representation of the Temperature (used, for example, in our display functions).

### LINKED ISSUE #12: Add Temperature::toStr()
Add a method to the Temperature class that will return a String (or char*) formatted appropriately for displaying on the Screen. If you're feeling fancy, make a second method that takes a parameter for the number of spaces after the decimal.

Once you've done that, refactor the various display methods so that they use the new toStr() method.

This would eliminate duplication of this sort of effort in all the class methods that display Temperature results on the Screen.

kyleoliveira:
Resolves #12 .


kyleoliveira:
@azyth Check this out and let me know if there's anything off with it.

PR #16, opened 2015-09-23
Gets the Publisher class working foreals, including an example in rhizome.ino (commented out).

kyleoliveira:
The Publisher class provides a standardized wrapper around Spark.publish so we don't have to think too hard about sending our messages to Ohmbrewer. It should be used at the appropriate stage of doWork() to send those messages. For example, in a Relay-based class, you'd want to use a Publisher to send Ohmbrewer an update after you toggle the relay (whether due to a state change or due to reaching the stop time).

### DIFF
--- a/firmware/rhizome.ino
+++ b/firmware/rhizome.ino
@@ -7,6 +7,7 @@
 #include "Ohmbrewer_Publisher.h"
 #include "Ohmbrewer_Thermostat.h"
 #include "Ohmbrewer_RIMS.h"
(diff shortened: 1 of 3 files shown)


kyleoliveira:
@azyth At some point (probably the next sprint), we'll want to write some publish() methods that will do the thinking necessary for standard updates from within doWork(). For now, I just figured I'd get this working since its fairly easy and make sure you're aware of it in case you want it for debugging purposes.

Don't worry about using one of these just yet if you don't want to.

Issue #17, opened 2015-09-28
Refactor pins implementation


kyleoliveira:
Due to questions raised in #8 , we've determined that we should refactor how we keep track of pins. Since pretty much all equipment needs at least two pins and probably no more than two pins, swap out the pin list and add in two new variables (including getters and setters) for a control pin and a state pin. These should still be ints.

Refactor whichPins to report the pins in use as well.

Additionally, it would probably be a good idea to create a very thin object between Equipment and the next level up instead of putting these pin variables in Equipment directly because RIMS and Thermostat do not have pins directly. Any arguments against that (besides perceived complexity)?

kyleoliveira:
Moved to #19

PR #18, opened 2015-09-29
equipment pin refactor #17

azyth:
changes up the variables in equipment

### DIFF
--- a/lib/Ohmbrewer_Equipment.h
+++ b/lib/Ohmbrewer_Equipment.h
@@ -33,6 +33,7 @@ namespace Ohmbrewer {
              */
             typedef std::map <String, String> args_map_t;
 
+
             /**
              * The Equipment ID
              * @returns The Sprout ID to use for this piece of Equipment
@@ -119,19 +120,17 @@ namespace Ohmbrewer {
             /**
              * Constructor
              * @param id The Sprout ID to use for this piece of Equipment
-             * @param pins The list of physical pins this Equipment is attached to
              */
-            Equipment(int id, std::list<int>* pins);
+            Equipment(int id,  int powerPin, int controlPin);
 
             /**
              * Constructor
              * @param id The Sprout ID to use for this piece of Equipment
-             * @param pins The list of physical pins this Equipment is attached to
              * @param stopTime The time at which the Equipment should shut off, assuming it isn't otherwise interrupted
              * @param state Whether the Equipment is ON (or OFF). True => ON, False => OFF
              * @param currentTask The unique identifier of the task that the Equipment believes it should be processing
              */
-            Equipment(int id, std::list<int>* pins, int stopTime, bool state, String currentTask);
+            Equipment(int id, int stopTime, bool state, String currentTask);
 
             /**
              * Copy Constructor
--- a/lib/Ohmbrewer_Relay.cpp
+++ b/lib/Ohmbrewer_Relay.cpp
@@ -4,9 +4,12 @@
 /**
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param pins The list of physical pins this Equipment is attached to
+ * @param powerPin - The power pin - on/off line. Digital pin number X.
+ * @param controlPin - The Control pin - Data/speed/power level Digital pin number X.
  */
-Ohmbrewer::Relay::Relay(int id, std::list<int>* pins) : Ohmbrewer::Equipment(id, pins) {
+Ohmbrewer::Relay::Relay(int id, int powerPin, int controlPin) : Ohmbrewer::Equipment(id) {
+    _powerPin = powerPin;
+    _controlPin = controlPin;
     // For now, we will not automatically add a Spark.function to Relays as
     // it's used mostly as a base class and our subclasses call the Relay constructor. If we can find a safe way to
     // determine if a function for actual Relay types should be added, then we'll change that.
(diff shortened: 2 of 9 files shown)


azyth:
TODO
remove pins from equipment 
transfer pins functions to temperature sensor (bus) 
and relay (power and control).

change contructors to match

lib/Ohmbrewer_Equipment.h
```diff
@@ -119,19 +120,17 @@ namespace Ohmbrewer {
             /**
              * Constructor
              * @param id The Sprout ID to use for this piece of Equipment
-             * @param pins The list of physical pins this Equipment is attached to
              */
-            Equipment(int id, std::list<int>* pins);
+            Equipment(int id,  int powerPin, int controlPin);
```
kyleoliveira, line 124: Equipment won't know about pins (aside from the whichPins interface-y method), so don't include this version of the constructor. Just drop it.

lib/Ohmbrewer_Relay.cpp
```diff
@@ -44,6 +50,54 @@ Ohmbrewer::Relay::~Relay() {
     // Nothing to do here...
```
kyleoliveira, line 52: Now there is something to do here...

lib/Ohmbrewer_Relay.cpp
```diff
@@ -44,6 +50,54 @@ Ohmbrewer::Relay::~Relay() {
...
+int Ohmbrewer::Relay::getPowerPin(){
+    return _powerPin;
+}
+
+/**
+ * Sets the Digital pin for the power pin.
+ * @param pinNum Dx
+ * @returns The time taken to run the method
+ */
+const int Ohmbrewer::Relay::setPowerPin(int pinNum) {
```
kyleoliveira: That parameter would be a const int, no?

lib/Ohmbrewer_Relay.cpp
```diff
@@ -44,6 +50,54 @@ Ohmbrewer::Relay::~Relay() {
     // Nothing to do here...
 }
 
+
+/**
+ * The power pin - on/off line
+ * hardware destinations (hardware switches - DPDT, main power)
+ * @returns The pin number in use for this piece of Equipment
+ */
+int Ohmbrewer::Relay::getPowerPin(){
```
kyleoliveira: In this case, the return value is a const int.

lib/Ohmbrewer_Relay.cpp
```diff
@@ -44,6 +50,54 @@ Ohmbrewer::Relay::~Relay() {
...
+    return start - millis();
+}
+
+
+/**
+ * The Control pin - Data/speed/power level
+ * hardware destinations (SSR, databus(temp sensor), pump speed control)
+ * @returns The pin number in use for this piece of Equipment
+ */
+int Ohmbrewer::Relay::getControlPin(){
```
kyleoliveira: You'll want to return a const int here.

lib/Ohmbrewer_Relay.cpp
```diff
@@ -44,6 +50,54 @@ Ohmbrewer::Relay::~Relay() {
...
+    return _controlPin;
+}
+
+/**
+ * Sets the Digital pin for the control pin.
+ * @param pinNum Dx
+ * @returns The time taken to run the method
+ */
+const int Ohmbrewer::Relay::setControlPin(int pinNum) {
```
kyleoliveira: I believe the parameter would be a const int here.

lib/Ohmbrewer_Relay.cpp
```diff
@@ -118,12 +172,31 @@ bool Ohmbrewer::Relay::isOff() const {
...
  * Performs the Equipment's current task. Expect to use this during loop().
+ * Switches the power pin on or off for a generic relay, subclasses can utilize the controlPin
  * This function is called by work().
  * @returns The time taken to run the method
  */
 int Ohmbrewer::Relay::doWork() {
-    // TODO: Implement Relay::doWork
-    return -1;
+    int startTime = millis();
+    bool relayState = getState();
```
kyleoliveira: There isn't any advantage to creating local variables for these things you're using getters for.

lib/Ohmbrewer_Temperature_Sensor.cpp
```diff
@@ -56,6 +58,7 @@ Ohmbrewer::TemperatureSensor::TemperatureSensor(int id, std::list<int>* pins, in
  * @param clonee The Equipment object to copy
  */
 Ohmbrewer::TemperatureSensor::TemperatureSensor(const TemperatureSensor& clonee) : Ohmbrewer::Equipment(clonee) {
+    _busPin = clonee.getBusPin;
```
kyleoliveira: You're going to need to add parenthesis to this function or you're gonna have a bad time.

lib/Ohmbrewer_Temperature_Sensor.h
```diff
@@ -78,10 +78,25 @@ namespace Ohmbrewer {
...
             virtual ~TemperatureSensor();
-            
+
+
+            /**
+             * The Bus pin - Data input line
+             * onewire protocol input location for DS18b20
+             * @returns The pin number in use for this piece of Equipment
+             */
+            int getBusPin();
```
kyleoliveira: getBusPin() should return a const int. Update here and in the .cpp, of course.

lib/Ohmbrewer_Temperature_Sensor.h
```diff
@@ -78,10 +78,25 @@ namespace Ohmbrewer {
...
+
             /**
-             * Overloaded << operator.
+             * Sets the Digital pin for the data Bus.
+             * @param pinNum Dx
+             * @returns The time taken to run the method
              */
+            const int setBusPin(int pinNum) ;
```
kyleoliveira: pinNum should be a const int.

lib/Ohmbrewer_Relay.cpp
azyth, line 52: is there, we still dont have any objects/memory that we have allocated in here

lib/Ohmbrewer_Equipment.h
azyth, line 124: i dropped this constructor to just ID

lib/Ohmbrewer_Relay.cpp
kyleoliveira, line 52: What I mean is, this comment isn't necessary anymore because of what you added on lines 41 and 42 (i.e. you're doing something here).

lib/Ohmbrewer_Relay.h
```diff
@@ -61,10 +63,39 @@ namespace Ohmbrewer {
...
              */
             virtual ~Relay();
-            
+
+            /**
+             * The power pin - on/off line
+             * hardware destinations (hardware switches - DPDT, main power)
+             * @returns The pin number in use for this piece of Equipment
+             */
+            int getPowerPin() const;
```
kyleoliveira, line 72: Still need a const on the left.

lib/Ohmbrewer_Relay.h
```diff
@@ -61,10 +63,39 @@ namespace Ohmbrewer {
...
+             * @returns The time taken to run the method
              */
+            const int setPowerPin(int pinNum);
+
+            /**
+             * The Control pin - Data/speed/power level
+             * hardware destinations (SSR, databus(temp sensor), pump speed control)
+             * @returns The pin number in use for this piece of Equipment
+             */
+            int getControlPin() const;
```
kyleoliveira, line 86: Still need a const on the left.

kyleoliveira:
See #19

PR #19, opened 2015-09-29
Refactors how pins work #17

### LINKED ISSUE #17: Refactor pins implementation

kyleoliveira:
Just squashed stuff from #18 (let's close that out without merge). Should resolve #17 .

### DIFF
--- a/lib/Ohmbrewer_Equipment.cpp
+++ b/lib/Ohmbrewer_Equipment.cpp
@@ -8,17 +8,15 @@ Ohmbrewer::Equipment::Equipment() {
     _id = 0;
     _stopTime = 0;
     _state = false;
-    _pins = new std::list<int>;
+
 }
 
 /**
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param pins The list of physical pins this Equipment is attached to
  */
-Ohmbrewer::Equipment::Equipment(int id, std::list<int>* pins) {
+Ohmbrewer::Equipment::Equipment(int id) {
     _id = id;
-    _pins = pins;
     _stopTime = 0;
     _state = false;
 }
@@ -26,14 +24,12 @@ Ohmbrewer::Equipment::Equipment(int id, std::list<int>* pins) {
 /**
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param pins The list of physical pins this Equipment is attached to
  * @param stopTime The time at which the Equipment should shut off, assuming it isn't otherwise interrupted
  * @param state Whether the Equipment is ON (or OFF). True => ON, False => OFF
  * @param currentTask The unique identifier of the task that the Equipment believes it should be processing
  */
-Ohmbrewer::Equipment::Equipment(int id, std::list<int>* pins, int stopTime, bool state, String currentTask) {
+Ohmbrewer::Equipment::Equipment(int id, int stopTime, bool state, String currentTask) {
     _id = id;
-    _pins = pins;
     _stopTime = stopTime;
     _state = state;
 }
@@ -44,7 +40,6 @@ Ohmbrewer::Equipment::Equipment(int id, std::list<int>* pins, int stopTime, bool
  */
 Ohmbrewer::Equipment::Equipment(const Equipment& clonee) {
     _id = clonee.getID();
-    _pins = clonee.whichPins();
     _stopTime = clonee.getStopTime();
     _state = clonee.getState();
     _currentTask = clonee.getCurrentTask();
@@ -223,7 +218,7 @@ int Ohmbrewer::Equipment::assignArgs(args_map_t &argsMap) {
  * Destructor
  */
 Ohmbrewer::Equipment::~Equipment() {
-    delete _pins;
+
 }
 
 /**
--- a/lib/Ohmbrewer_Equipment.h
+++ b/lib/Ohmbrewer_Equipment.h
@@ -14,6 +14,7 @@
 #include <list>
 #include "application.h"
 
+
 namespace Ohmbrewer {
     // Forward declaration
     class Screen;
@@ -32,6 +33,7 @@ namespace Ohmbrewer {
              * mapped to the appropriate Equipment members.
              */
             typedef std::map <String, String> args_map_t;
+
 
             /**
              * The Equipment ID
@@ -119,19 +121,17 @@ namespace Ohmbrewer {
             /**
              * Constructor
              * @param id The Sprout ID to use for this piece of Equipment
-             * @param pins The list of physical pins this Equipment is attached to
              */
-            Equipment(int id, std::list<int>* pins);
+            Equipment(int id);
 
             /**
              * Constructor
              * @param id The Sprout ID to use for this piece of Equipment
-             * @param pins The list of physical pins this Equipment is attached to
              * @param stopTime The time at which the Equipment should shut off, assuming it isn't otherwise interrupted
              * @param state Whether the Equipment is ON (or OFF). True => ON, False => OFF
              * @param currentTask The unique identifier of the task that the Equipment believes it should be processing
              */
-            Equipment(int id, std::list<int>* pins, int stopTime, bool state, String currentTask);
+            Equipment(int id, int stopTime, bool state, String currentTask);
 
             /**
              * Copy Constructor
@@ -215,7 +215,7 @@ namespace Ohmbrewer {
              * Reports which of the Rhizome's pins are occupied by the
              * Equipment, forming a logical Sprout.
              */
-            virtual std::list<int>* whichPins() const = 0;
+            virtual void whichPins(std::list<int>* pins) = 0;
 
         protected:
             /**
@@ -232,12 +232,6 @@ namespace Ohmbrewer {
              * State (True => On, False => Off)
              */
             bool           _state;
-
-            /**
-             * Which of the Rhizome's pins are occupied by the
-             * Equipment, forming a logical Sprout.
-             */
-            std::list<int>* _pins;
 
             /**
              * The Task the Equipment is currently processing.
--- a/lib/Ohmbrewer_Heating_Element.cpp
+++ b/lib/Ohmbrewer_Heating_Element.cpp
@@ -3,22 +3,26 @@
 /**
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param pins The list of physical pins this Equipment is attached to
+ * @param powerPin - The power pin - on/off line. Digital pin number X.
+ * @param controlPin - The Control pin - Data/speed/power level Digital pin number X.
  */
-Ohmbrewer::HeatingElement::HeatingElement(int id, std::list<int>* pins) : Ohmbrewer::Relay(id, pins) {
+Ohmbrewer::HeatingElement::HeatingElement(int id, int powerPin, int controlPin) : Ohmbrewer::Relay(id, powerPin, controlPin) {
+
     registerUpdateFunction();
 }
 
 /**
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param pins The list of physical pins this Equipment is attached to
+ * @param powerPin - The power pin - on/off line. Digital pin number X.
+ * @param controlPin - The Control pin - Data/speed/power level Digital pin number X.
  * @param stopTime The time at which the Equipment should shut off, assuming it isn't otherwise interrupted
  * @param state Whether the Equipment is ON (or OFF). True => ON, False => OFF
  * @param currentTask The unique identifier of the task that the Equipment believes it should be processing
  */
-Ohmbrewer::HeatingElement::HeatingElement(int id, std::list<int>* pins, int stopTime,
-                                          bool state, String currentTask) : Ohmbrewer::Relay(id, pins, stopTime, state, currentTask) {
+Ohmbrewer::HeatingElement::HeatingElement(int id, int powerPin, int controlPin, int stopTime,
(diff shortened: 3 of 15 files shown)


azyth:
there are still a few instances of the int list in existance but they are linked to output of which pins so didnt want to change those yet,

lib/Ohmbrewer_Equipment.cpp
```diff
@@ -223,7 +218,7 @@ int Ohmbrewer::Equipment::assignArgs(args_map_t &argsMap) {
  * Destructor
  */
 Ohmbrewer::Equipment::~Equipment() {
-    delete _pins;
+    //delete _pins;
```
kyleoliveira: You should completely remove this instead of just commenting out.

lib/Ohmbrewer_Heating_Element.cpp
```diff
@@ -3,22 +3,28 @@
 /**
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param pins The list of physical pins this Equipment is attached to
+ * @param powerPin - The power pin - on/off line. Digital pin number X.
+ * @param controlPin - The Control pin - Data/speed/power level Digital pin number X.
  */
-Ohmbrewer::HeatingElement::HeatingElement(int id, std::list<int>* pins) : Ohmbrewer::Relay(id, pins) {
+Ohmbrewer::HeatingElement::HeatingElement(int id, int powerPin, int controlPin) : Ohmbrewer::Relay(id, powerPin, controlPin) {
+//    _powerPin = powerPin;
```
kyleoliveira: You'll want to clear out these commented lines.

lib/Ohmbrewer_Heating_Element.cpp
```diff
@@ -3,22 +3,28 @@
...
  * @param stopTime The time at which the Equipment should shut off, assuming it isn't otherwise interrupted
  * @param state Whether the Equipment is ON (or OFF). True => ON, False => OFF
  * @param currentTask The unique identifier of the task that the Equipment believes it should be processing
  */
-Ohmbrewer::HeatingElement::HeatingElement(int id, std::list<int>* pins, int stopTime,
-                                          bool state, String currentTask) : Ohmbrewer::Relay(id, pins, stopTime, state, currentTask) {
+Ohmbrewer::HeatingElement::HeatingElement(int id, int powerPin, int controlPin, int stopTime,
+                                          bool state, String currentTask) : Ohmbrewer::Relay(id, powerPin, controlPin, stopTime, state, currentTask) {
+//    _powerPin = powerPin;
+//    _controlPin = controlPin;
```
kyleoliveira: You'll want to clear out these commented lines.

lib/Ohmbrewer_Heating_Element.cpp
```diff
@@ -28,6 +34,8 @@ Ohmbrewer::HeatingElement::HeatingElement(int id, std::list<int>* pins, int stop
  */
 Ohmbrewer::HeatingElement::HeatingElement(const HeatingElement& clonee) : Ohmbrewer::Relay(clonee) {
     // This has probably already been set, but maybe clonee is a more complicated child class...
+//    _powerPin = clonee.getPowerPin();
+//    _controlPin = clonee.getControlPin();
```
kyleoliveira: More comments to remove.

lib/Ohmbrewer_Heating_Element.cpp
```diff
@@ -38,6 +46,52 @@ Ohmbrewer::HeatingElement::~HeatingElement() {
     // Nothing to do here...
 }
 
+//
```
kyleoliveira: Remove this commented out section and any commented out references to stuff below.

lib/Ohmbrewer_Heating_Element.h
```diff
@@ -67,10 +69,39 @@ namespace Ohmbrewer {
              * Destructor
              */
             virtual ~HeatingElement();
-            
-            /**
-             * Overloaded << operator.
-             */
+
```
kyleoliveira: Remove this commented stuff.

lib/Ohmbrewer_RIMS.cpp
```diff
@@ -28,57 +28,57 @@ Ohmbrewer::Pump* Ohmbrewer::RIMS::getRecirculator() const {
...
+ * @param tunBus - mash tun temperature pin
+ * @param pumpPins[powerPin ; controlPin ]
  */
-Ohmbrewer::RIMS::RIMS(int id, std::list<int>* pins) : Ohmbrewer::Equipment(id, pins) {
-    // TODO: Figure out how to properly set the components of the RIMS the constructors
-    std::list<int>* fakePins = new std::list<int>(1,-1);
-    _tube = new Thermostat(1, fakePins);
-    _tunSensor = new TemperatureSensor(1, fakePins);
-    _recirc = new Pump(1,fakePins);
+Ohmbrewer::RIMS::RIMS(int id, int tubePins[], int tunBus, int pumpPins[] ) : Ohmbrewer::Equipment(id) {
```
kyleoliveira: You should have a check to ensure that the arrays that are passed in are appropriately sized. If it fails, maybe initialize with the ID-only constructor and/or send an error out using the Publisher class?

lib/Ohmbrewer_RIMS.cpp
```diff
@@ -28,57 +28,57 @@ Ohmbrewer::Pump* Ohmbrewer::RIMS::getRecirculator() const {
...
+    _tube = new Thermostat(1, tubePins);
+    _tunSensor = new TemperatureSensor(1, tunBus);
```
kyleoliveira: Assuming the TemperatureSensor is created after the Thermostat as is is here, and that there are no other TemperatureSensors in existence on this Rhizome, the ID would be '2' and not '1'. You'll want to fix this in other constructors too.

lib/Ohmbrewer_RIMS.cpp
```diff
@@ -359,7 +359,11 @@ int Ohmbrewer::RIMS::doUpdate(String &args, Ohmbrewer::Equipment::args_map_t &ar
  */
 std::list<int>* Ohmbrewer::RIMS::whichPins() const {
     // TODO: Implement RIMS::whichPins
+
+    //TODO return array of pins
```
kyleoliveira: You should implement whichPins() as part of this PR... If you fix up the return statement on line 366 (and implement whichPins() properly in the getters it uses), you should be pretty much there.

One unfortunate thing is that whichPins should probably be more like:
`void whichPins(std::list<int>* pinList)`
or
`void whichPins(int* pinList)`
where pinList is a buffer created in the calling function. This is because if you create the pinList within the method itself you'll either get a blank pointer back or (possibly) create a memory leak. See Temperature::toStrC as an example using a character string. It isn't any more complicated to do it which a std::list, and the list would concatenate more easily.

lib/Ohmbrewer_Relay.cpp
```diff
@@ -99,12 +152,25 @@ bool Ohmbrewer::Relay::isOff() const {
...
 /**
  * Performs the Equipment's current task. Expect to use this during loop().
+ * Switches the power pin on or off for a generic relay, subclasses can utilize the controlPin
  * This function is called by work().
  * @returns The time taken to run the method
  */
 int Ohmbrewer::Relay::doWork() {
-    // TODO: Implement Relay::doWork
-    return -1;
+    int startTime = millis();
```
kyleoliveira: If this commit has your doWork work in it and you merge it into your current branch for getting doWork working, you might have a bad time... Just a warning. You might want to make this part of the work more atomic. Your call.

If you do want to roll this part back on this branch (assuming it exists on the other branch), CLion make it easier to revert this kind of thing.

lib/Ohmbrewer_Heating_Element.h
azyth: this was yours, i didnt know what it was for

kyleoliveira: No, I'm talking about the commented out pin stuff. Since you've put the pin stuff in Relay and TemperatureSensor, any references that are commented out (like the getPowerPin method below), should be cleaned up.

lib/Ohmbrewer_Equipment.h
```diff
@@ -237,7 +237,7 @@ namespace Ohmbrewer {
              * Which of the Rhizome's pins are occupied by the
              * Equipment, forming a logical Sprout.
              */
-            std::list<int>* _pins;
+            //std::list<int>* _pins;
```
kyleoliveira: Remove this commented out stuff.

lib/Ohmbrewer_Heating_Element.cpp
```diff
@@ -3,22 +3,28 @@
...
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param pins The list of physical pins this Equipment is attached to
+ * @param powerPin - The power pin - on/off line. Digital pin number X.
+ * @param controlPin - The Control pin - Data/speed/power level Digital pin number X.
  */
-Ohmbrewer::HeatingElement::HeatingElement(int id, std::list<int>* pins) : Ohmbrewer::Relay(id, pins) {
+Ohmbrewer::HeatingElement::HeatingElement(int id, int powerPin, int controlPin) : Ohmbrewer::Relay(id, powerPin, controlPin) {
+//    _powerPin = powerPin;
+//    _controlPin = controlPin;
```
kyleoliveira: Remove this commented-out stuff above.

lib/Ohmbrewer_Heating_Element.cpp
kyleoliveira: Remove this commented-out stuff above

lib/Ohmbrewer_Heating_Element.cpp
kyleoliveira: Remove this commented-out stuff above

lib/Ohmbrewer_Heating_Element.cpp
```diff
@@ -50,6 +104,7 @@ Ohmbrewer::HeatingElement::~HeatingElement() {
  */
 int Ohmbrewer::HeatingElement::doWork() {
     // TODO: Implement HeatingElement::doWork
+    //digitalWrite( someWayOfGettingTheSinglePin(), (getState() : HIGH ? LOW) )
```
kyleoliveira: Remove this commented-out stuff above or you're probably going to have a bad time merging, etc.

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -383,8 +439,11 @@ int Ohmbrewer::Thermostat::doUpdate(String &args, Ohmbrewer::Equipment::args_map
...
 std::list<int>* Ohmbrewer::Thermostat::whichPins() const {
-    // TODO: Implement Thermostat::whichPins
-    //return {&_pins, &getElement(), &getSensor()};
-    return _pins;
+    //return {temp buPin, element powerPin, element controlPin};
+    std::list<int>* pins = new std::list<int>();
+    pins->push_back(_tempSensor->getBusPin());
+    pins->push_back(_heatingElm->getPowerPin());
+    pins->push_back(_heatingElm->getControlPin());
+    return pins;
```
kyleoliveira: Instead of manually specifying the pins for the heating element, just concatenate the list to what you've pushed into pins.

See the comment in RIMS above.

lib/Ohmbrewer_RIMS.cpp
```diff
@@ -358,8 +404,23 @@ int Ohmbrewer::RIMS::doUpdate(String &args, Ohmbrewer::Equipment::args_map_t &ar
...
+//    pins = _tube->whichPins();
+//    pins->push_back(&(_tunSensor->getBusPin()));
+//    pins->push_back(&(_recirc->getPowerPin()));
+//    pins->push_back(&(_recirc->getControlPin()));
+
+    // [ tube temp busPin ; tube heating powerPin ; tube heating controlPin ; mashtun temp busPin ; pump powerPin ; pump controlPin]
+
     //return {&_pins, &getTube()->whichPins(), &getTunSensor()->whichPins(), &getRecirculator()->whichPins()};
-    return _pins;
+    return pins;
```
kyleoliveira: Instead of manually specifying the pins for the tube and recirculation pump, just concatenate the list to what you've already pushed into pins.
See http://stackoverflow.com/questions/1449703/how-to-append-a-listt-object-to-another for details.

i.e. 

```
tubePins = _tube->whichPins();
recircPins = _recirc->whichPins();
pins->insert(pins.end(), tubePins.begin(), tubePins.end());
//etc.
```

The advantage here is that you don't have to keep specifying the order (consistently, every time), and if for some strange reason the pin variables, etc. for the tube were to change, you don't have to update RIMS.

lib/Ohmbrewer_Heating_Element.cpp
azyth: pretty sure this one is from the other PR as is

kyleoliveira: That's kinda my point... It hasn't been merged to master yet. So we're going to merge this PR and it'll get in master. Then you're going to work on the previous PR and it's probably going to say you've got a conflict because this line was changed in both or something (especially if you rebase to get the new stuff from master into that existing PR).

PR #23, opened 2015-10-10
Dallas redux - adds support for the dallas onewire sensor

azyth:
@kyleoliveira got the squash to work right but all the other merged commits are still sitting on top


azyth:
In the future I think that i cannot try to merge/rebase before i squash my commits. I am open to suggestions on how not to run into this problem.

kyleoliveira:
This ain't quite right. We'll try again...

Issue #25, opened 2015-10-12
Refactor RIMS sensor layout


kyleoliveira:
After thinking about it a bit, we've come to the conclusion that the sensor layout in the RIMS configuration doesn't make much sense with respect to how we want to run our PID algorithm. Namely, the PID algorithm should be contained within the Thermostat and in order for it to work properly, it needs to know the temperature of the mash - not the temperature in the tube. Hence,

1) Fix the names of getters/setters within RIMS so that the TemperatureSensor that will track temperature in the tun is the one within the RIMS object's Thermostat.
2) Fix the names of getters/setters within RIMS so that the TemperatureSensor that will track temperature in the tube is the non-Thermostat sensor for the RIMS. Name it something like "safety sensor" or "check sensor" as it is intended to provide the RIMS with a warning for when the tube is reaching unacceptable heating limits.
3) Provide delegating getters/setters within RIMS for both sensors for ease of use in display functions and so forth.
4) Double-check that you've fixed up display methods, update methods, and constructors so that things are "wired" correctly.

azyth:
Been thinking about this, and it may be pretty simple, mainly thermostat will want a "safety sensor".  we can keep most things the same and just add a few getter/setter type things in thermostat.

and switching which sensor is set as the Therm temp sensor (needs to be mash sensor, not tube sensor)

kyleoliveira:
That seems reasonable to me. Go for it.

Issue #9, opened 2015-09-19
Implement doWork for Thermostat


kyleoliveira:
Depends on #7 and #8 ...

Once you have the TemperatureSensor and HeatingElement classes' doWork methods in place, you should be able to implement the PID functionality in Thermostat's doWork method.

kyleoliveira:
Probably want to finish #25 before you get going on this one...

Issue #8, opened 2015-09-19
Implement doWork for Relay and its subclasses


kyleoliveira:
This shouldn't be too hard... I'm not even sure there's anything to do other than have doWork return a sensible value. Since Relay and its subclasses are simply ON/OFF switches and the switching occurs in doUpdate, that's probably it.

Analyze, confirm, and implement as appropriate.

kyleoliveira:
Thought about this a little bit, and it is probably appropriate for doWork to do the actual switching on the pin based on what is set in the _state variable (as opposed to doing so in doUpdate upon receipt of an update). In other words, doWork should do something like:

```
  digitalWrite( someWayOfGettingTheSinglePin(), (getState() : HIGH ? LOW) );
```

Still very simple, but doing it in doWork should make for a nice separation of concerns / clean interface.

azyth:
@kyleoliveira So the part of this that gets tricky is with the modulation of the elements with the SSRs and switches.

in come cases we may need to control one element with two pins...pump also(later) one for on off and one for speed/heat level. if we have _pins as a list supplied to us via equipment. whichpins(). we still do not have a way to distinguish which pins are which. do we?

one way around this would be to just use two SSRs and they could be switched in parallel fro both sides of the phase.  but using a switch is usually cheaper but they cannot handle constant switching since they will wear out.

kyleoliveira:
So, there are probably a number of ways to deal with that issue, but the first that come to my mind are:
1) The pins are supplied in an ordered list, so we can simply require that the first pin supplied is the on/off and the second handles the speed
2) Explicitly break whichPins (and probably the pin storage mechanism) into separate variables.

The laziest/easiest is probably to stick with (1).

However, (2) might not be a bad idea... if we did that, we'd probably remove references to "pins" from the Equipment class and handle it in the next level up (Relay and TemperatureSensor). Each of those would need one pin for On/Off. Then the next level up from that (HeatingElement and Pump) wouldn't need anything new (beyond the initial refactoring of Equipment). Finally, when we add support for the next level up from that (variable speed pumps and variable heat elements), those classes would have a variable for the pin to use for handling variability.

Another reason why (2) is kind of a good idea is because of the question of how to handle pin settings for composite Equipment types (Thermostat and RIMS). Really those classes shouldn't care about pins - they're only around to represent a logical grouping of equipment, so including a list of pins is kind of silly (but required at the moment due to the interface defined by Equipment).

[...]

PR #29, opened 2015-10-19
Thermostat PID #25 #9

azyth:
All in one PR for both #25 and #9  and #8 .  

relay doWork() completed for basic on off functionality of single switch relay
sensor refactoring to account for RIMS and Still

doWork() once sensor structure in place.

### DIFF
--- a/firmware/rhizome.ino
+++ b/firmware/rhizome.ino
@@ -7,14 +7,15 @@
 #include "Ohmbrewer_Publisher.h"
 #include "Ohmbrewer_Thermostat.h"
 #include "Ohmbrewer_RIMS.h"
+#include "Ohmbrewer_Onewire.h"
 #include "onewire.h"
-// #include "Ohmbrewer_Publisher.h" // (*)
 
 // Kludge to allow us to use std::list - for now we have to undefine these macros.
 #undef min
 #undef max
 #undef swap
 #include <deque>
+#include <list>
 
 /* ========================================================================= */
 /*  Global Vars                                                              */
@@ -31,6 +32,9 @@ std::deque< Ohmbrewer::Equipment* > sprouts;
  * update it within the Sprout management functions.
  */
 int inUse[6] = {0,0,0,0,0,0};
+
+//initializer Pin list
+std::list<int> thermPins (1, 0);
 
 /**
  * The touchscreen object. Handles the display for the Rhizome.
@@ -51,7 +55,7 @@ unsigned long lastUpdate = millis();
  * Does any preliminary setup work before the Rhizome starts the operation loop.
  */
 void setup() {
-    // Serial.begin(9600); // Enable serial for debugging messages
+//    Serial.begin(9600); // Enable serial for debugging messages
     String fakeTask = "fake";
 
     ow_setPin(D0); //This should later be accomplished by equipment setup OR constructor
@@ -59,37 +63,34 @@ void setup() {
     // Add our initial Equipment. We wouldn't necessarily do this, but it's useful for now.
     // We'll also set some temperatures and relay values explicitly when we probably wouldn't.
 // EX 1: Temperature Sensors
-//    sprouts.push_back(new Ohmbrewer::TemperatureSensor( 1, new std::deque<int>(1,0) ));
+//    sprouts.push_back(new Ohmbrewer::TemperatureSensor( 1, new std::list<int>(1,0) ));
 //    ((Ohmbrewer::TemperatureSensor*)sprouts.back())->getTemp()->set(42);
 //
-//    sprouts.push_back(new Ohmbrewer::TemperatureSensor( 2, new std::list<int>(1,1) ));
+//    sprouts.push_back(new Ohmbrewer::TemperatureSensor( 2, new std::list<int>(1,0) ));
 //    ((Ohmbrewer::TemperatureSensor*)sprouts.back())->getTemp()->set(-20);
 
 // EX 2: A Thermostat
-//    sprouts.push_back(new Ohmbrewer::Thermostat( 1, new std::list<int>(1,0), 100 ));
+//    sprouts.push_back(new Ohmbrewer::Thermostat( 1, new std::list<int>(3,0), 100 ));
 //    ((Ohmbrewer::Thermostat*)sprouts.front())->getElement()->setState(true);
 
 // EX 3: Pumps & Heating Elements
 //    sprouts.push_back(new Ohmbrewer::Pump( 1, new std::list<int>(1,2), 0, true, fakeTask ));
 //    sprouts.push_back(new Ohmbrewer::Pump( 2, new std::list<int>(1,3) ));
-//    sprouts.push_back(new Ohmbrewer::HeatingElement( 1, new std::list<int>(1,4) ));
-//    sprouts.push_back(new Ohmbrewer::HeatingElement( 2, new std::list<int>(1,5), 0, true, fakeTask ));
+//    sprouts.push_back(new Ohmbrewer::HeatingElement( 3, new std::list<int>(1,4) ));
+//    sprouts.push_back(new Ohmbrewer::HeatingElement( 4, new std::list<int>(1,5), 0, true, fakeTask ));
 
 // EX 4: A RIMS w/ a Pump to the Kettle
-    //TODO change to array
-    // [ tube temp busPin ; tube heating powerPin ; tube heating controlPin ; mashtun temp busPin ; pump powerPin ; pump controlPin]
-
-    int tubePins[] = {0,1,2};
-    int tunPin = 3;
-    int pumpPins[] = {4,5};
-    //for(int i = 1; i < 5; i++) {rimsPins->push_back(i);}
-
-    sprouts.push_back(new Ohmbrewer::RIMS( 1, tubePins, tunPin, pumpPins ));
+    thermPins.push_back(2); //control pin (relay pin)
+    thermPins.push_back(3); //power pin (switch pin)
+    int pumpPin = 4;
+    int pumpPin2 = 5;
+    sprouts.push_back(new Ohmbrewer::RIMS( 1, &thermPins, pumpPin ));
+    sprouts.push_back(new Ohmbrewer::Pump( 8 , pumpPin2 )); //transfer pump
     ((Ohmbrewer::RIMS*)sprouts.front())->setState(true); // Turn everything on.
-    ((Ohmbrewer::RIMS*)sprouts.front())->getTube()->setTargetTemp(559);
+    ((Ohmbrewer::Thermostat*)sprouts.front())->setState(true); // Turn everything on.
+    ((Ohmbrewer::RIMS*)sprouts.front())->getTube()->setTargetTemp(62.7);
 
-    sprouts.push_back(new Ohmbrewer::Pump( 2, pumpPins[0], pumpPins[1] ));
-
+//TURN ON screen
     screen.initScreen();
 //    pMap[String("hey")] = String("listen!"); // (*)
 }
@@ -98,19 +99,23 @@ void setup() {
  * The meat of the program. Runs repeatedly until the Rhizome is powered off.
  */
 void loop() {
+    //probe ID finder command (switch this for the other below commands to enable)
+//    ((Ohmbrewer::Onewire*)(((Ohmbrewer::RIMS*)sprouts.front())->getTube()->getSensor()->getProbe()))
+//            ->displayProbeIds(&screen);
+
+
+
     if((millis() - lastUpdate) > 10000) {
         // Toggle the last relay every 10 seconds, for illustration.
-//        sprouts.back()->setState(sprouts.back()->isOff()); // The last HeatingElement
-//        ((Ohmbrewer::Thermostat*)sprouts.at(0))->getElement()
-//                                               ->setState(!((Ohmbrewer::Thermostat*)sprouts.at(0))->getElement()
-//                                                                                                  ->getState()); // The Thermostat's HeatingElement
         ((Ohmbrewer::RIMS*)sprouts.front())->getTube()
                                            ->getElement()
                                            ->toggleState();
-//        navi->publish(); // (*)
+
         lastUpdate = millis();
     }
-    ((Ohmbrewer::RIMS*)sprouts.front())->getTube()->getSensor()->doWork(); //Temp patch to make sensor read in TUBE
+    ((Ohmbrewer::RIMS*)sprouts.front())->getTube()->getSensor()->work(); //Temp patch to make sensor read in Tun temp
+
+    //refresh the display
     screen.refreshDisplay();
 }
 
--- a/lib/Ohmbrewer_Heating_Element.cpp
+++ b/lib/Ohmbrewer_Heating_Element.cpp
@@ -1,12 +1,14 @@
 #include "Ohmbrewer_Heating_Element.h"
+#include "Ohmbrewer_Screen.h"
 
 /**
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param powerPin - The power pin - on/off line. Digital pin number X.
- * @param controlPin - The Control pin - Data/speed/power level Digital pin number X.
+ * @param elementPins - controlPin always first in <list>
+ *  controlPin - The Control pin - Data/speed/power level Digital pin number X.
+ *  powerPin - The power pin - on/off line. Digital pin number X.
  */
-Ohmbrewer::HeatingElement::HeatingElement(int id, int powerPin, int controlPin) : Ohmbrewer::Relay(id, powerPin, controlPin) {
+Ohmbrewer::HeatingElement::HeatingElement(int id, std::list<int>* elementPins) : Ohmbrewer::Relay(id, elementPins) {
 
     registerUpdateFunction();
 }
@@ -14,14 +16,15 @@ Ohmbrewer::HeatingElement::HeatingElement(int id, int powerPin, int controlPin) 
 /**
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param powerPin - The power pin - on/off line. Digital pin number X.
- * @param controlPin - The Control pin - Data/speed/power level Digital pin number X.
+ * @param elementPins - controlPin always first in <list>
+ *  controlPin - The Control pin - Data/speed/power level Digital pin number X.
+ *  powerPin - The power pin - on/off line. Digital pin number X.
  * @param stopTime The time at which the Equipment should shut off, assuming it isn't otherwise interrupted
  * @param state Whether the Equipment is ON (or OFF). True => ON, False => OFF
  * @param currentTask The unique identifier of the task that the Equipment believes it should be processing
  */
-Ohmbrewer::HeatingElement::HeatingElement(int id, int powerPin, int controlPin, int stopTime,
-                                          bool state, String currentTask) : Ohmbrewer::Relay(id, powerPin, controlPin, stopTime, state, currentTask) {
+Ohmbrewer::HeatingElement::HeatingElement(int id, std::list<int>* elementPins, int stopTime,
+                                          bool state, String currentTask) : Ohmbrewer::Relay(id, elementPins, stopTime, state, currentTask) {
 
     registerUpdateFunction();
 }
@@ -42,23 +45,13 @@ Ohmbrewer::HeatingElement::HeatingElement(const HeatingElement& clonee) : Ohmbre
 Ohmbrewer::HeatingElement::~HeatingElement() {
 }
 
-
-/**
- * Performs the Equipment's current task. Expect to use this during loop().
- * This function is called by work().
- * @returns The time taken to run the method
- */
-int Ohmbrewer::HeatingElement::doWork() {
-    // TODO: Implement HeatingElement::doWork
-    return -1;
-}
-
 /**
  * Draws information to the Rhizome's display.
  * This function is called by display().
  * @returns The time taken to run the method
  */
 int Ohmbrewer::HeatingElement::doDisplay(Ohmbrewer::Screen *screen) {
+    screen->print("HeatE.");
     return Relay::doDisplay(screen);
 }
 
--- a/lib/Ohmbrewer_Onewire.cpp
+++ b/lib/Ohmbrewer_Onewire.cpp
@@ -3,16 +3,17 @@
 #include "ds18x20.h"
 #include "onewire.h"
 
+
 /**
  * Constructors
  * @param busPin Digital pin that serves as the bus for all onewire
  * @param probeId Unique ID for the temperature probe [8] char array ID code
  */
 Ohmbrewer::Onewire::Onewire() : Ohmbrewer::Probe(){
-//    ow_setPin(D0);
+//    ow_setPin(D0);    //set globally
 }
 
-Ohmbrewer::Onewire::Onewire(char* probeId) : Ohmbrewer::Probe(){
+Ohmbrewer::Onewire::Onewire(uint8_t (&probeId)[8]) : Ohmbrewer::Probe(){
 //    ow_setPin(D0);
     _probeId = probeId;
 }
@@ -72,7 +73,7 @@ double Ohmbrewer::Onewire::getReading(){
         if (sensors[i*OW_ROMCODE_SIZE+0] == 0x10 || sensors[i*OW_ROMCODE_SIZE+0] == 0x28) //0x10=DS18S20, 0x28=DS18B20
         {
             // if current probe matches the probe we are looking for
-            if ( strcmp(_probeId, probeId) ){
+            if ( strcmp(_probeId, probeId) ){ //TODO does this work for a hex value?
                 if ( DS18X20_read_meas( &sensors[0], &subzero, &cel, &celFracBits) == DS18X20_OK ) {
                     char sign = (subzero) ? '-' : '+';
                     int frac = celFracBits*DS18X20_FRACCONV;
@@ -99,17 +100,90 @@ double Ohmbrewer::Onewire::getReading(){
 }
 
 /**
+ * outputs probe IDs and their current temperatures to the screen
+ *
+ * Primarily for use as a manual identification tool for probe IDs
+ *
+ * FIXME probeID output is currently in HEX something or other, *non readable*
+ */
+void Ohmbrewer::Onewire::displayProbeIds(Ohmbrewer::Screen *screen){
+    //header
+    screen->resetTextSize();
+    screen->resetTextColor();
+    // Print the section title
+    screen->print("PROBE ID# : TEMP");
+//    screen->println("");
+    screen->printMargin(2);
+
+
+    uint8_t sensors[80];
+    uint8_t subzero, cel, celFracBits;        //local vars
+    double tempC;
+    //Asks all DS18x20 devices to start temperature measurement, takes up to 750ms at max resolution
+    DS18X20_start_meas( DS18X20_POWER_PARASITE, NULL );
+    //If your code has other tasks, you can store the timestamp instead and return when a second has passed.
+    delay(1000);
+    uint8_t numSensors = ow_search_sensors(10, sensors);
+    screen->displayHeader();
+    for (uint8_t i=0; i<numSensors; i++){
+        //current probe ID
+        char probeId[8] = {
+                sensors[(i*OW_ROMCODE_SIZE)+0],
+                sensors[(i*OW_ROMCODE_SIZE)+1],
+                sensors[(i*OW_ROMCODE_SIZE)+2],
+                sensors[(i*OW_ROMCODE_SIZE)+3],
+                sensors[(i*OW_ROMCODE_SIZE)+4],
+                sensors[(i*OW_ROMCODE_SIZE)+5],
+                sensors[(i*OW_ROMCODE_SIZE)+6],
+                sensors[(i*OW_ROMCODE_SIZE)+7]};
+
+        if (sensors[i*OW_ROMCODE_SIZE+0] == 0x10 || sensors[i*OW_ROMCODE_SIZE+0] == 0x28){ //0x10=DS18S20, 0x28=DS18B20
+           //for each probe print probe and temp to screen
+            if ( DS18X20_read_meas( &sensors[0], &subzero, &cel, &celFracBits) == DS18X20_OK ) {
+                //char sign = (subzero) ? '-' : '+';
+                int frac = celFracBits*DS18X20_FRACCONV;
+                tempC = (double)cel;
+                tempC = tempC + (.0001*(double)frac);
+
+                //print the ID and temp TODO fix me! hex value?
+                screen->print(probeId);
+//                screen->print(probeId[0]);
+//                screen->print(probeId[1]);
+//                screen->print(probeId[2]);
+//                screen->print(probeId[3]);
+//                screen->print(probeId[4]);
+//                screen->print(probeId[5]);
+//                screen->print(probeId[6]);
+//                screen->print(probeId[7]);
+                screen->print(":");
+                screen->println(tempC);
+
+            }
+            else {
+                if (i == numSensors - 1) {
+                    //error
+                    tempC = -69;
+                    screen->print(probeId);
+                    screen->print(":");
+                    screen->println(tempC);
+                }
+            }
+        }
+    }
+}
+
+/**
  * sets the probe id for this instance
  * @param probe id  Unique probe ID
  */
-void Ohmbrewer::Onewire::setProbeId(char* probeId){
+void Ohmbrewer::Onewire::setProbeId(uint8_t (&probeId)[8]){
     _probeId = probeId;
 }
 
 /**
  * @returns _probeID Unique probe ID
  */
-char* Ohmbrewer::Onewire::getProbeId(){
+uint8_t* Ohmbrewer::Onewire::getProbeId(){
     return _probeId;
 }
 
--- a/lib/Ohmbrewer_Onewire.h
+++ b/lib/Ohmbrewer_Onewire.h
@@ -4,6 +4,7 @@
 
 #include "Ohmbrewer_Probe.h"
 #include "application.h"
+#include "Ohmbrewer_Screen.h"
 
 
 namespace Ohmbrewer {
@@ -17,7 +18,7 @@ namespace Ohmbrewer {
          */
         Onewire();
 
-        Onewire(char* probeId);
+        Onewire(uint8_t (&probeId)[8]);
 
         /**
          * @returns the Celsius reading from the specified connected DS18b20 probe
@@ -25,17 +26,23 @@ namespace Ohmbrewer {
          */
         double getReading();
 
+        /**
+         * outputs probe IDs and their current temperatures to the screen
+         *
+         * Primarily for use as a manual identification tool for probe IDs
+         */
+        void displayProbeIds(Screen *screen);
 
         /**
          * sets the probe id for this instance
          * @param probe id  Unique probe ID
          */
-        void setProbeId(char* probeId);
+        void setProbeId(uint8_t (&probeId)[8]);
 
         /**
          * @returns _probeID Unique probe ID
          */
-        char* getProbeId();
+        uint8_t* getProbeId();
 
         /**
          * reads current probes connected to busPin
@@ -54,7 +61,7 @@ namespace Ohmbrewer {
         /**
          * Unique ID for the temperature probe [8] char array ID code
          */
-        char* _probeId;
+        uint8_t* _probeId;
 
         /**
          * Digital pin that the data stream is on
--- a/lib/Ohmbrewer_Pump.cpp
+++ b/lib/Ohmbrewer_Pump.cpp
@@ -1,26 +1,28 @@
 #include "Ohmbrewer_Pump.h"
+#include "Ohmbrewer_Publisher.h"
+#include "Ohmbrewer_Screen.h"
 
 /**
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param powerPin - The power pin - on/off line. Digital pin number X.
- * @param controlPin - The Control pin - Data/speed/power level Digital pin number X.
+ * @param pumpPin - Single speed pump will only have PowerPin
  */
-Ohmbrewer::Pump::Pump(int id, int powerPin, int controlPin) : Ohmbrewer::Relay(id, powerPin, controlPin) {
+Ohmbrewer::Pump::Pump(int id, int pumpPin) : Ohmbrewer::Relay(id, pumpPin) {
+
     registerUpdateFunction();
 }
 
-/**
+/**FIXME
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param powerPin - The power pin - on/off line. Digital pin number X.
- * @param controlPin - The Control pin - Data/speed/power level Digital pin number X.
+ * @param pumpPin - Single speed pump will only have PowerPin
  * @param stopTime The time at which the Equipment should shut off, assuming it isn't otherwise interrupted
  * @param state Whether the Equipment is ON (or OFF). True => ON, False => OFF
  * @param currentTask The unique identifier of the task that the Equipment believes it should be processing
  */
-Ohmbrewer::Pump::Pump(int id, int powerPin, int controlPin, int stopTime,
-                      bool state, String currentTask) : Ohmbrewer::Relay(id, powerPin, controlPin, stopTime, state, currentTask) {
+Ohmbrewer::Pump::Pump(int id, int pumpPin, int stopTime,
+                      bool state, String currentTask) : Ohmbrewer::Relay(id, pumpPin, stopTime, state, currentTask) {
+
     registerUpdateFunction();
 }
 
@@ -61,6 +63,7 @@ int Ohmbrewer::Pump::doWork() {
  * @returns The time taken to run the method
  */
 int Ohmbrewer::Pump::doDisplay(Ohmbrewer::Screen *screen) {
+    screen->print("Pump");
     return Relay::doDisplay(screen);
 }
 
--- a/lib/Ohmbrewer_Pump.cpp
+++ b/lib/Ohmbrewer_Pump.cpp
@@ -1,26 +1,28 @@
 #include "Ohmbrewer_Pump.h"
+#include "Ohmbrewer_Publisher.h"
+#include "Ohmbrewer_Screen.h"
 
 /**
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param powerPin - The power pin - on/off line. Digital pin number X.
- * @param controlPin - The Control pin - Data/speed/power level Digital pin number X.
+ * @param pumpPin - Single speed pump will only have PowerPin
  */
-Ohmbrewer::Pump::Pump(int id, int powerPin, int controlPin) : Ohmbrewer::Relay(id, powerPin, controlPin) {
+Ohmbrewer::Pump::Pump(int id, int pumpPin) : Ohmbrewer::Relay(id, pumpPin) {
+
     registerUpdateFunction();
 }
 
-/**
+/**FIXME
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param powerPin - The power pin - on/off line. Digital pin number X.
- * @param controlPin - The Control pin - Data/speed/power level Digital pin number X.
+ * @param pumpPin - Single speed pump will only have PowerPin
  * @param stopTime The time at which the Equipment should shut off, assuming it isn't otherwise interrupted
  * @param state Whether the Equipment is ON (or OFF). True => ON, False => OFF
  * @param currentTask The unique identifier of the task that the Equipment believes it should be processing
  */
-Ohmbrewer::Pump::Pump(int id, int powerPin, int controlPin, int stopTime,
-                      bool state, String currentTask) : Ohmbrewer::Relay(id, powerPin, controlPin, stopTime, state, currentTask) {
+Ohmbrewer::Pump::Pump(int id, int pumpPin, int stopTime,
+                      bool state, String currentTask) : Ohmbrewer::Relay(id, pumpPin, stopTime, state, currentTask) {
+
     registerUpdateFunction();
 }
 
@@ -56,6 +58,7 @@ int Ohmbrewer::Pump::doWork() {
  * @returns The time taken to run the method
  */
 int Ohmbrewer::Pump::doDisplay(Ohmbrewer::Screen *screen) {
+    screen->print("Pump");
     return Relay::doDisplay(screen);
 }
 
--- a/lib/Ohmbrewer_RIMS.cpp
+++ b/lib/Ohmbrewer_RIMS.cpp
@@ -4,6 +4,97 @@
 
 
 /**
+ * Constructor - creating a RIMS will use 7 equipment IDs
+ * @param id The Sprout ID to use for this piece of Equipment
+ * @param thermPins list with formatting of: [ temp busPin ;  heating controlPin ; heating powerPin ]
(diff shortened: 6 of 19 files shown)


azyth:
@kyleoliveira what do you think about the safety sensor idea?  Thermostat will need access to both but not sure if there would be a better way to do it. (still rought draft at this point)

azyth:
Relay dowork completed for basic on off functionality #8

lib/Ohmbrewer_Onewire.cpp
```diff
@@ -2,6 +2,7 @@
 #include "Ohmbrewer_Onewire.h"
 #include "ds18x20.h"
 #include "onewire.h"
+#include <string>
```
kyleoliveira: Use Particle Strings instead of the C++ String class.

lib/Ohmbrewer_Onewire.cpp
```diff
@@ -99,6 +100,85 @@ double Ohmbrewer::Onewire::getReading(){
...
+                    tempC = -69;
+                    //screen->print("%s:%d", probeId, tempC);
+                    screen->print(probeId[0]);
+                    screen->print(probeId[1]);
+                    screen->print(probeId[2]);
+                    screen->print(probeId[3]);
+                    screen->print(probeId[4]);
+                    screen->print(probeId[5]);
+                    screen->print(probeId[6]);
+                    screen->print(probeId[7]);
```
kyleoliveira: Does

``` c++
screen->print(probeId);
// or
screen->print(probeId, HEX); // a guess
```

not get the job done?

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -1,52 +1,36 @@
 #include "Ohmbrewer_Thermostat.h"
 #include "Ohmbrewer_Screen.h"
 #include "Ohmbrewer_Publisher.h"
+#include "pid.h"
 
+//Define Variables we'll be connecting to with PID
+double Setpoint, Input, Output;
```
kyleoliveira: Camelcase, yo.

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -1,52 +1,36 @@
...
-/**
- * The Thermostat's heating element
- * @returns The heating element
- */
-Ohmbrewer::HeatingElement* Ohmbrewer::Thermostat::getElement() const {
-    return _heatingElm;
-}
+//PID windowing variables
+int WindowSize = 5000;
+unsigned long windowStartTime;
```
kyleoliveira: Also, why are all these variables in lines 6-15 not included under the class? Doing this puts them into the global scope...

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -263,9 +331,56 @@ bool Ohmbrewer::Thermostat::isOff() const {
...
+        myPID.SetTunings(aggKp, aggKi, aggKd);
+    }
+    myPID.Compute();
+
+    // turn the output pin on/off based on pid output
+    if (millis() - windowStartTime>WindowSize) { //time to shift the Relay Window
+        windowStartTime += WindowSize;
+    }
+
+    if (Output < millis() - windowStartTime) {
```
kyleoliveira: Where does Output get set?

lib/Ohmbrewer_Onewire.cpp
azyth: I will try it, i was didnt get to far into debuggin that yet.

lib/Ohmbrewer_Onewire.cpp
azyth: not sure i will even need it

azyth:
Constructors throughout the firmware are now TODO, I will go through and patch them before this is finished but waiting to make sure everything is set before replicating work....again. 
basic constructor is the only functional one at the moment for most classes.

went back to <list> (yes you where right it is easier) so that constructors will be simpler for single speed pumps/ relays with/without switches etc.... 

PID functionality MAY have longer timeouts then we would like since it is not the only thing running in loop{}, but real time windowing should help with that.

I used alot of default values from the pid library, could probably use tweaking with testing.
testing the PID dowork stuff is upcomming. 

Probe UIDs are stiill WIP, not technically part of this PR, but if it works without too much extra fiddling I will include here since it is imporatant to the PID stuff. If not I will start a new PR for just that once this is closed. 

Currently ONLY ONE temp probe is supported per pin, and OW only supports one global pin atm also. so safety temp sensor functionality is not implemented yet.

firmware/rhizome.ino
```diff
@@ -7,7 +7,9 @@
 #include "Ohmbrewer_Publisher.h"
 #include "Ohmbrewer_Thermostat.h"
 #include "Ohmbrewer_RIMS.h"
+#include "Ohmbrewer_Onewire.h"
 #include "onewire.h"
+
```
kyleoliveira, line 12: Should probably remove this extra space :wink:

lib/Ohmbrewer_Onewire.cpp
```diff
@@ -99,6 +100,87 @@ double Ohmbrewer::Onewire::getReading(){
...
+//                    screen->print(probeId[3]);
+//                    screen->print(probeId[4]);
+//                    screen->print(probeId[5]);
+//                    screen->print(probeId[6]);
+//                    screen->print(probeId[7]);
+                    screen->print(":");
+                    screen->print(tempC);
+                }
+            }
+        }
```
kyleoliveira, line 171: You probably want to add a

``` c++
screen->println("");
```

hereish so each sensor prints on a separate line.

lib/Ohmbrewer_Relay.cpp
```diff
@@ -1,14 +1,33 @@
...
+ */
+Ohmbrewer::Relay::Relay(int id, std::list<int>* relayPins) : Ohmbrewer::Equipment(id) {
+    //FIXME
+    int size = relayPins->size();
+    int powerPin = -69;
+    int controlPin;
+    if  (size > 1){ // switch and relay
+        controlPin = relayPins->front();
+        relayPins->pop_front();
+        powerPin = relayPins->front();
```
kyleoliveira: Given that the size is 2 (or at least you can expand that if to verify as much), you can use back() and not have to do any popping. There is a small, unlikely chance that whoever passed these relayPins in might want to continue using the list.

lib/Ohmbrewer_Onewire.cpp
```diff
@@ -99,6 +100,100 @@ double Ohmbrewer::Onewire::getReading(){
...
+ *
+ * FIXME probeID output is currently in HEX something or other, *non readable*
+ */
+void Ohmbrewer::Onewire::displayProbeIds(Ohmbrewer::Screen *screen){
+    //header
+    screen->resetTextSize();
+    screen->resetTextColor();
+    // Print the section title
+    screen->print("PROBE ID# : TEMP");
+    screen->println("");
```
kyleoliveira: Why not just merge lines 114 and 115?

``` c++
screen->println("PROBE ID# : TEMP");
```

lib/Ohmbrewer_Pump.cpp
```diff
@@ -1,16 +1,31 @@
...
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param powerPin - The power pin - on/off line. Digital pin number X.
- * @param controlPin - The Control pin - Data/speed/power level Digital pin number X.
+ * powerPin - The power pin - on/off line. Digital pin number X. (switch)
+ * controlPin - The Control pin - Data/speed/power level Digital pin number X. (relay) TODO
+ * @param pumpPins - controlPin is always first in <list> single speed pump will only have PowerPin however
  */
-Ohmbrewer::Pump::Pump(int id, int powerPin, int controlPin) : Ohmbrewer::Relay(id, powerPin, controlPin) {
+Ohmbrewer::Pump::Pump(int id, std::list<int>* pumpPins) : Ohmbrewer::Relay(id) {
```
kyleoliveira: Can this be delegated down to Relay?

lib/Ohmbrewer_Temperature_Sensor.h
```diff
@@ -35,35 +35,19 @@ namespace Ohmbrewer {
...
+            TemperatureSensor(int id, int busPin);
 
             /**
              * Constructor
              * @param id The Sprout ID to use for this TemperatureSensor
              * @param pins The list of physical pins this TemperatureSensor is attached to
+             * @param uid - the unique id code for the Dallas sensor connected to the bus.
              */
-            TemperatureSensor(int id, int busPin);
+            TemperatureSensor(int id, int busPin, char* uid);
```
kyleoliveira: If you know how long the UID is (8, right?), you can limit the size as we've done in other parameters.

lib/Ohmbrewer_Pump.cpp
azyth: no. because of how pumps are controlled as opposed to relays,

azyth: actualy I think i just thought of a way

lib/Ohmbrewer_Onewire.cpp
azyth: because i wasnt sure if i needed the second line to actually space it out.

firmware/rhizome.ino
```diff
@@ -76,19 +80,28 @@ void setup() {
...
+    int pumpPin = 3;
+//    int pumpPin2 = 4;
+//    thermPins->pop_front();
 
-    sprouts.push_back(new Ohmbrewer::RIMS( 1, tubePins, tunPin, pumpPins ));
-    ((Ohmbrewer::RIMS*)sprouts.front())->setState(true); // Turn everything on.
-    ((Ohmbrewer::RIMS*)sprouts.front())->getTube()->setTargetTemp(559);
+//    sprouts.push_back(new Ohmbrewer::RIMS( 1, thermPins, pumpPin ));          //DOSNT WORK
+    sprouts.push_back(new Ohmbrewer::Thermostat(1, thermPins));                 //WORKS
```
azyth: @kyleoliveira  these two lines above, if you switch them on one it works the other it hardfaults. looking through the constructors I cannot find a place where I do anything substantially differently. please take a look and tell me if you see anything.

lib/Ohmbrewer_RIMS.cpp
```diff
@@ -3,58 +3,37 @@
...
-        if ( n > 1){
-        _recirc = new Pump(1, pumpPins[0], pumpPins[1]);
-    }else if (n == 1){
-        _recirc = new Pump(1, pumpPins[0], -1);//Single speed pump
+Ohmbrewer::RIMS::RIMS(int id, std::list<int>* thermPins, int pumpPin ) : Ohmbrewer::Equipment(id) {
+    //TODO currently only one probe at a time working.
+    int size = thermPins->size();
+    int tempBus;
+    if ( (size == 2) || (size == 3) ){
+        tempBus = thermPins->front();
```
kyleoliveira: Why make an extra variable?

firmware/rhizome.ino
```diff
@@ -7,6 +7,7 @@
 #include "Ohmbrewer_Publisher.h"
 #include "Ohmbrewer_Thermostat.h"
 #include "Ohmbrewer_RIMS.h"
+#include "Ohmbrewer_Onewire.h"
 #include "onewire.h"
```
kyleoliveira: You probably also want to add an ``` #include <list> ``` on line 19. The only reason it isn't yelling about this must be that we include list within the equipment.

firmware/rhizome.ino
```diff
@@ -32,6 +33,9 @@ std::deque< Ohmbrewer::Equipment* > sprouts;
  */
 int inUse[6] = {0,0,0,0,0,0};
 
+std::list<int>* thermPins = new std::list<int>(1,0);
```
kyleoliveira: Change this to:

``` c++
std::list<int> thermPins (1, 0);
```

or, alternatively, you can load up all the pins in 2 lines (like a boss):

``` c++
int thmP[] = {0,4,5};
std::list<int> thermPins (thmP, thmP + sizeof(thmP) / sizeof(int));
```

firmware/rhizome.ino
kyleoliveira: If you change thermPins as I advise above (i.e. change it from a list pointer to simply a list), then this works:

``` c++
    sprouts.push_back(new Ohmbrewer::RIMS( 1, &thermPins, pumpPin )); //DOES WORK
    sprouts.push_back(new Ohmbrewer::Thermostat(1, &thermPins));          //WORKS

```

firmware/rhizome.ino
```diff
@@ -98,20 +111,32 @@ void setup() {
...
+//    ((Ohmbrewer::Onewire*)(((Ohmbrewer::RIMS*)sprouts.front())->getTube()->getSensor()->getProbe()))
+//            ->displayProbeIds(&screen);
+//    screen.println( ((Ohmbrewer::HeatingElement*)sprouts.front())->getControlPin() );
+//    screen.println( ((Ohmbrewer::HeatingElement*)sprouts.front())->getPowerPin() );
+// int a = 0;
+//    if (a==0){
+//        sprouts.push_back(new Ohmbrewer::Thermostat(1, thermPins));
+//        a++;
+//    }
+    screen.println(thermPins->front());
```
kyleoliveira: I commented this out, just in case. Seems like it _shouldn't_ break anything, though...

firmware/rhizome.ino
```diff
@@ -98,20 +111,32 @@ void setup() {
...
+//        ((Ohmbrewer::RIMS*)sprouts.front())->getTube()
+//                                           ->getElement()
+//                                           ->toggleState();
 //        navi->publish(); // (*)
         lastUpdate = millis();
     }
-    ((Ohmbrewer::RIMS*)sprouts.front())->getTube()->getSensor()->doWork(); //Temp patch to make sensor read in TUBE
-    screen.refreshDisplay();
+//    ((Ohmbrewer::RIMS*)sprouts.front())->getTube()->getSensor()->work(); //Temp patch to make sensor read in TUBE
+//    screen.refreshDisplay();
```
kyleoliveira: I brought this line back so I could see what's going on (assuming things don't hard fault, of course).

firmware/rhizome.ino
azyth: this was just a test to make sure the list was passed throguh the constructors and popped correctly

firmware/rhizome.ino
azyth: then is it really nessasary?

lib/Ohmbrewer_RIMS.cpp
azyth: initially I think i used this var more.

azyth:
everything should be up to date unless i overlooked something.  going to get it merge-able before i resolve the merge conflicts with master

firmware/rhizome.ino
kyleoliveira: Until such time that we don't use an std::list in rhizome.ino, I'd put it in there. It doesn't hurt, at the very least.

lib/Ohmbrewer_Heating_Element.cpp
```diff
@@ -43,15 +45,16 @@ Ohmbrewer::HeatingElement::~HeatingElement() {
 }
 
 
-/**
+/**TODO remove overwriting?
```
kyleoliveira: Yes, if HeatingElement always does what Relay does for a method, just leave the method out of HeatingElement. We can always add it back in, if needed.

lib/Ohmbrewer_Onewire.cpp
```diff
@@ -3,16 +3,18 @@
...
 
+
 /**
  * Constructors
  * @param busPin Digital pin that serves as the bus for all onewire
  * @param probeId Unique ID for the temperature probe [8] char array ID code
  */
 Ohmbrewer::Onewire::Onewire() : Ohmbrewer::Probe(){
-//    ow_setPin(D0);
+    Serial.println("Onewire");
```
kyleoliveira: You can clear out the debugging statements now that it's working (or at least comment them out).

lib/Ohmbrewer_Onewire.cpp
```diff
@@ -3,16 +3,18 @@
...
+//    ow_setPin(D0);    //set globally
```
kyleoliveira, line 13: Looking at this PR, I'm wondering if it would be a good idea to make a static function in Ohmbrewer::Onewire that just wraps ow_setPin(). It would mean we don't have to include onewire.h in the .ino file and it would be a little cleaner when we implement other Probe types. Something like Ohmbrewer::Onewire::init(int busProbe) or some such. You'd call it in the exact same place in the .ino.
Not strictly necessary for this PR though. Just throwing it out there.

lib/Ohmbrewer_Onewire.cpp
```diff
@@ -99,17 +101,111 @@ double Ohmbrewer::Onewire::getReading(){
...
+                //print the ID and temp TODO fix me! hex value?
+                screen->print(probeId);
+//                screen->print(probeId[0]);
+//                screen->print(probeId[1]);
+//                screen->print(probeId[2]);
+//                screen->print(probeId[3]);
+//                screen->print(probeId[4]);
+//                screen->print(probeId[5]);
+//                screen->print(probeId[6]);
+//                screen->print(probeId[7]);
```
kyleoliveira, line 157: Clear out 151-158 if it isn't needed.

lib/Ohmbrewer_Onewire.cpp
```diff
@@ -99,17 +101,111 @@ double Ohmbrewer::Onewire::getReading(){
...
+                    //screen->print("%s:%d", probeId, tempC);
+                    screen->print(probeId);
+//                    screen->print(probeId[0]);
+//                    screen->print(probeId[1]);
+//                    screen->print(probeId[2]);
+//                    screen->print(probeId[3]);
+//                    screen->print(probeId[4]);
+//                    screen->print(probeId[5]);
+//                    screen->print(probeId[6]);
+//                    screen->print(probeId[7]);
```
kyleoliveira: Clear out 170-177 if it isn't needed.

lib/Ohmbrewer_Onewire.cpp
```diff
@@ -99,17 +101,111 @@ double Ohmbrewer::Onewire::getReading(){
...
+//                screen->print(probeId[1]);
+//                screen->print(probeId[2]);
+//                screen->print(probeId[3]);
+//                screen->print(probeId[4]);
+//                screen->print(probeId[5]);
+//                screen->print(probeId[6]);
+//                screen->print(probeId[7]);
+                screen->print(":");
+                screen->print(tempC);
+                screen->println("");
```
kyleoliveira: It looks like you print the temperature row the same way regardless of how the if statement lands. You could make a single method that gets called in both places.

lib/Ohmbrewer_Onewire.cpp
```diff
@@ -99,17 +101,111 @@ double Ohmbrewer::Onewire::getReading(){
...
+//                    screen->print(probeId[1]);
+//                    screen->print(probeId[2]);
+//                    screen->print(probeId[3]);
+//                    screen->print(probeId[4]);
+//                    screen->print(probeId[5]);
+//                    screen->print(probeId[6]);
+//                    screen->print(probeId[7]);
+                    screen->print(":");
+                    screen->print(tempC);
+                    screen->println("");
```
kyleoliveira: Instead of the empty println(), change the previous print to a println. All println does is do a print and automatically append a newline.

lib/Ohmbrewer_Onewire.h
```diff
@@ -25,17 +26,30 @@ namespace Ohmbrewer {
...
+         * Primarily for use as a manual identification tool for probe IDs
+         */
+        void displayProbeIds(Screen *screen);
+
+        /**
+         * assigns the probe ID to supplied onewire probe
+         * @param probe id  Unique probe ID
+         * @param probe  - probe
+         */
+        void assignProbeId(Ohmbrewer::Onewire probe, uint8_t (&probeId)[8]);//TODO
```
kyleoliveira: It looks like you want this to be a static method. You're trying to make something that isn't associated with a particular instance of the class, right?

Otherwise, I don't see a whole lot of advantage of this over the method it's delegating to (setProbeId).
Even then, if you know the probe already, what's the point of this?

lib/Ohmbrewer_Pump.cpp
```diff
@@ -61,6 +63,7 @@ int Ohmbrewer::Pump::doWork() {
  * @returns The time taken to run the method
  */
 int Ohmbrewer::Pump::doDisplay(Ohmbrewer::Screen *screen) {
+    screen->print("Pump");
```
kyleoliveira, line 61: I don't think you want this.

lib/Ohmbrewer_Onewire.cpp
azyth, line 157: I am not sure exactly what is needed for this yet, going to resolve Unique ID stuff in a seperate PR

lib/Ohmbrewer_RIMS.cpp
```diff
@@ -4,120 +4,89 @@
...
+        _tube = new Thermostat(id+3, thermPins, targetTemp);
+        _tunSensor = new TemperatureSensor(id+1, tempBus);
     }else{
         //publish error
         Ohmbrewer::Publisher::publish_map_t pMap;
         Ohmbrewer::Publisher* pub = new Ohmbrewer::Publisher(new String("error_log"), &pMap);
-
-        pMap[String("array_check_rims")] = String("improperly formed array - RIMS(int , int[], int, int[], int, bool, String, double)");
+        pMap[String("list_check_rims")] = String("improperly formed array - RIMS(int, int<list>, int");
```
kyleoliveira: Side note: We should make another Publisher constructor that takes a single string such that you can do lines 84-85 in a single line. That seem slike it would be handy.

lib/Ohmbrewer_RIMS.cpp
```diff
@@ -4,120 +4,89 @@
...
+    int size = thermPins->size();
+    int tempBus;
+    if ( (size == 2) || (size == 3) ){
+        tempBus = thermPins->front();
+        _tube = new Thermostat(id+3, thermPins, targetTemp);
+        _tunSensor = new TemperatureSensor(id+1, tempBus);
     }else{
         //publish error
         Ohmbrewer::Publisher::publish_map_t pMap;
         Ohmbrewer::Publisher* pub = new Ohmbrewer::Publisher(new String("error_log"), &pMap);
```
kyleoliveira: I don't think you need to create this as a pointer. If you do / insist on this, you need to delete the pointer when you're done or we're going to get memory leaks.

lib/Ohmbrewer_RIMS.cpp
```diff
@@ -198,7 +191,7 @@ void Ohmbrewer::RIMS::parseArgs(const String &argsStr, Ohmbrewer::Equipment::arg
  */
 const int Ohmbrewer::RIMS::setState(const bool state) {
     unsigned long start = millis();
-
+//    setState(state); infinitely recursive?  //Equipment::setState(state);?????
```
kyleoliveira: Yup. Instead, do this:

``` c++
_state = state
```

lib/Ohmbrewer_RIMS.cpp
```diff
@@ -237,6 +230,9 @@ bool Ohmbrewer::RIMS::isOff() const {
  */
 int Ohmbrewer::RIMS::doWork() {
     // TODO: Implement RIMS::doWork
+    // enable thermotat?
+    //call them dowork()
+    //futz with pump a bit too.
```
kyleoliveira: Gonna leave this part to another PR, or just knock it out here?

lib/Ohmbrewer_RIMS.h
```diff
@@ -105,6 +85,24 @@ namespace Ohmbrewer {
...
+             * The Tube thermostat
+             * @returns The Thermostat object representing the RIMS tube elements
+             */
+            Thermostat* getTube() const;
+
+            /**
+             * The temperature sensor located in the tun
+             * @returns The Temperature Sensor object representing the sensor located in the mash tun
+             */
+            TemperatureSensor* getTunSensor() const;
```
kyleoliveira: Isn't this getSafetySensor now?

lib/Ohmbrewer_Relay.cpp
```diff
@@ -1,36 +1,89 @@
...
  * @param id The Sprout ID to use for this piece of Equipment
- * @param powerPin - The power pin - on/off line. Digital pin number X.
- * @param controlPin - The Control pin - Data/speed/power level Digital pin number X.
+ * @param pumpPin - Single speed pump will only have PowerPin
  * @param stopTime The time at which the Equipment should shut off, assuming it isn't otherwise interrupted
  * @param state Whether the Equipment is ON (or OFF). True => ON, False => OFF
  * @param currentTask The unique identifier of the task that the Equipment believes it should be processing
  */
-Ohmbrewer::Relay::Relay(int id, int powerPin, int controlPin, int stopTime,
+Ohmbrewer::Relay::Relay(int id, int pumpPin, int stopTime,
```
kyleoliveira: Should name the argument "powerPin" since that's what it is setting in the Relay.

lib/Ohmbrewer_Relay.cpp
```diff
@@ -1,36 +1,89 @@
...
-    // it's used mostly as a base class and our subclasses call the Relay constructor. If we can find a safe way to
-    // determine if a function for actual Relay types should be added, then we'll change that.
 }
 
 /**
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param powerPin - The power pin - on/off line. Digital pin number X.
- * @param controlPin - The Control pin - Data/speed/power level Digital pin number X.
+ * @param pumpPin - Single speed pump will only have PowerPin
```
kyleoliveira: Along with the comment below, fix this to something like "the power pin to use" or some such.

lib/Ohmbrewer_Onewire.h
azyth: i had a reason before but i cant think of exactly why i wanted a static method now

lib/Ohmbrewer_Relay.cpp
```diff
@@ -171,6 +249,7 @@ int Ohmbrewer::Relay::doDisplay(Ohmbrewer::Screen *screen) {
     char relay_id[2];
 
     // Print a fancy identifier
+    screen->print("ID#");
```
kyleoliveira, line 251: I don't think this is necessary.

lib/Ohmbrewer_Pump.cpp
azyth, line 61: yes I do, i refactored the display

lib/Ohmbrewer_Temperature_Sensor.cpp
```diff
@@ -49,11 +37,10 @@ Ohmbrewer::TemperatureSensor::TemperatureSensor(int id, int busPin) : Ohmbrewer:
...
 Ohmbrewer::TemperatureSensor::TemperatureSensor(int id,  int busPin, int stopTime, bool state, String currentTask) : Ohmbrewer::Equipment(id, stopTime, state, currentTask) {
-    _probe = new Onewire();
-    _lastReading = new Temperature(0);
-    _lastReadTime = Time.now();
-    registerUpdateFunction();
-
+//    _probe = new Onewire();
+//    _lastReading = new Temperature(-69);
+//    _lastReadTime = Time.now();
+//    registerUpdateFunction();
```
kyleoliveira: Should this stuff be uncommented?

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -1,93 +1,99 @@
...
-
-    _targetTemp->set(targetTemp);
+//PID windowing variables
+int windowSize = 5000;              //changeable value
+unsigned long windowStartTime;
 
-    return start - millis();
-}
+//PID set up
+PID* thermPID;
```
kyleoliveira: thermPID should be set up kind of like what we did with Probe instead of floating around in the global space. You don't need to make a wrapper around it, but you should provide it as a member of Thermostat and provide getters, setters, and mentions in constructors/destructors.

lib/Ohmbrewer_RIMS.cpp
azyth: seperate PR

lib/Ohmbrewer_RIMS.h
azyth: not in RIMS, in rims it stays as tun sensor, and actually safety sensor is the other one anyways

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -1,93 +1,99 @@
...
- * Sets the target temperature
- * @param targetTemp The new target temperature in Celsius
- * @returns The time taken to run the method
- */
-const int Ohmbrewer::Thermostat::setTargetTemp(const double targetTemp) {
-    unsigned long start = millis();
-
-    _targetTemp->set(targetTemp);
+//PID windowing variables
+int windowSize = 5000;              //changeable value
```
kyleoliveira: Since the values for these will be particular to a given Thermostat, all of these PID variables should be provided as members of Thermostat, with getters and setters. You can still set default values, but it would happen in Thermostat.h.

I'm assuming this stuff is here for now for debugging, though, right? :wink:

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -1,93 +1,99 @@
...
-        _tempSensor = new TemperatureSensor(id+1, thermPins[0]);
-    }else if (n == 2 ){                     // no switch, only power% (SSR in parallel or always on other leg)
-        _heatingElm = new HeatingElement(id+2, thermPins[1], -1);
-        _tempSensor = new TemperatureSensor(id+1, thermPins[0]);
-    }else{
-        //publish error
-
+Ohmbrewer::Thermostat::Thermostat(int id, std::list<int>* thermPins) : Ohmbrewer::Equipment(id) {
+    //PID set up
+    thermPID = (new PID(&input, &output, &setPoint, consKp, consKi, consKd, PID::DIRECT));
```
kyleoliveira: Will we always use PID::DIRECT, or is that a mode or some such that should be included in the member variables for Thermostat?

kyleoliveira: Also, you don't need the extra parentheses around this 'new' call.

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -1,93 +1,99 @@
...
+    //turn the PID on
+    thermPID->SetMode(PID::AUTOMATIC);
+
+    //initilize equipment components
+    int size = thermPins->size();
+    int tempB;
+    if ( (size == 2) || (size == 3) ){
+
+        tempB = thermPins->front();
+        thermPins->pop_front();
```
kyleoliveira: I don't think you should pop the list. Instead, pass in a subset.

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -1,93 +1,99 @@
...
-    int n = sizeof(thermPins) / sizeof(int);
-    if ( n > 2){
-        _heatingElm = new HeatingElement(id+2, thermPins[1], thermPins[2]);
-        _tempSensor = new TemperatureSensor(id+1, thermPins[0]);
-    }else if (n == 2 ){                     // no switch, only power% (SSR in parallel or always on other leg)
-        _heatingElm = new HeatingElement(id+2, thermPins[1], -1);
-        _tempSensor = new TemperatureSensor(id+1, thermPins[0]);
-    }else{
-        //publish error
+Ohmbrewer::Thermostat::Thermostat(int id, std::list<int>* thermPins, const double targetTemp) : Ohmbrewer::Equipment(id) {
```
kyleoliveira, line 23: All the stuff I mentioned for the previous constructor applies there too, naturally.

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -97,58 +103,81 @@ Ohmbrewer::Thermostat::Thermostat(int id, int (&thermPins)[3], const double targ
...
-    int n = sizeof(thermPins) / sizeof(int);
-    if ( n > 2){
-        _heatingElm = new HeatingElement(id+2, thermPins[1], thermPins[2]);
-        _tempSensor = new TemperatureSensor(id+1, thermPins[0]);
-    }else if (n == 2 ){                     // no switch, only power% (SSR in parallel or always on other leg)
-        _heatingElm = new HeatingElement(id+2, thermPins[1], -1);
-        _tempSensor = new TemperatureSensor(id+1, thermPins[0]);
-    }else{
-        //publish error
+    //PID set up
```
kyleoliveira: Ditto.

If you're doing the same thing in all of these, maybe you should make a `initPID` method and call it from the constructors?

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -97,58 +103,81 @@ Ohmbrewer::Thermostat::Thermostat(int id, int (&thermPins)[3], const double targ
...
  * @param id The Sprout ID to use for this piece of Thermostat
- * @param tubePins[ temp busPin ; heating powerPin ; heating controlPin ]
+ * @param thermPins list with formatting of: [ temp busPin ;  heating controlPin ; heating powerPin ]
  * @param stopTime The time at which the Thermostat should shut off, assuming it isn't otherwise interrupted
  * @param state Whether the Thermostat is ON (or OFF). True => ON, False => OFF
  * @param currentTask The unique identifier of the task that the Thermostat believes it should be processing
  * @param targetTemp The new target temperature in Celsius
  */
-Ohmbrewer::Thermostat::Thermostat(int id, int (&thermPins)[3], int stopTime,
+Ohmbrewer::Thermostat::Thermostat(int id, std::list<int>* thermPins, int stopTime,
```
kyleoliveira: See above.

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -227,9 +340,10 @@ void Ohmbrewer::Thermostat::parseArgs(const String &argsStr, Ohmbrewer::Equipmen
  */
 const int Ohmbrewer::Thermostat::setState(const bool state) {
     unsigned long start = millis();
-
+//    setState(state);
```
kyleoliveira: Either do what I put for this in RIMS, or, given that getState always returns the status of the subsystems, we can ignore the state variable for both and completely remove it from setState

If you want to do the latter, it would be good to initialize _state to NULL so any fool who tries to query it directly has a bad time.

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -239,7 +353,7 @@ const int Ohmbrewer::Thermostat::setState(const bool state) {
  * @returns True => On, False => Off
  */
 bool Ohmbrewer::Thermostat::getState() const {
-    return (getElement()->getState() || getSensor()->getState());
+    return (getElement()->getState() || getSensor()->getState() || getSafetySensor()->getState());
```
kyleoliveira: Thermostat only has one sensor. Maybe you meant to put this in RIMS?

lib/Ohmbrewer_RIMS.h
kyleoliveira: Isn't this getSafetySensor now?

Remember we've decided that the sensor within the Tube object shall actually be placed in the thing it's trying to control the temperature of (the mash tun) while the second sensor (the one that's actually inline with the tube) shall be the Safety Sensor.

I know this is a little confusing since we've named the Thermostat object within the RIMS "Tube" - if you can think of a better name ("Controller" maybe?), I'm down.

lib/Ohmbrewer_Relay.cpp
azyth, line 251: otherwise you might think it is the Pin #

lib/Ohmbrewer_Thermostat.cpp
kyleoliveira: Alternatively, if you're just trying to capture the default values to hand in to the PID object _and_ you're going to have a PID object in the Thermostat anyhow, you can set those values up as either static class variables or a struct. For the former, check out how I declared the named colors in Screen. Given that you have sets of stuff, maybe two std::map<char*, int> 's? One for aggressivePID one for conservativePID? Or simply two int[3]'s and the order is fixed/known.

lib/Ohmbrewer_Thermostat.cpp
azyth: i like popping !

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -339,12 +506,49 @@ unsigned long Ohmbrewer::Thermostat::displayCurrentTemp(Ohmbrewer::Screen *scree
 }
 
 /**
+ * Prints the temperature information for our safety sensor onto the touchscreen.
+ * @returns Time it took to run the function
+ */
+unsigned long Ohmbrewer::Thermostat::displaySafetyTemp(Ohmbrewer::Screen *screen) {
```
kyleoliveira: This goes in RIMS.

lib/Ohmbrewer_Thermostat.cpp
azyth: THermostat dosnt only have one sensor

lib/Ohmbrewer_Thermostat.h
```diff
@@ -113,6 +87,64 @@ namespace Ohmbrewer {
...
+             * @returns The temperature sensor
+             */
+            TemperatureSensor* getSafetySensor() const;
+
+            /**
+             * Sets the Thermostat's Safety temperature sensor
+             * @param sensor -  The temperature sensor
+             * @returns The time taken to run the method
+             */
+            const int setSafetySensor(TemperatureSensor* sensor);
```
kyleoliveira: Safety Sensor stuff goes in RIMS

lib/Ohmbrewer_Onewire.cpp
kyleoliveira, line 157: Well, we know we can at least print it like this (even though the character is wonky). I think if you change it to 

``` c++
screen->print(probeId, HEX);
```

it _may_ work. Even if it doesn't though, you're not going to use the commented out stuff and it should go.

lib/Ohmbrewer_Pump.cpp
kyleoliveira, line 61: The method in Screen that displays all the pumps (Screen::displayPumps()) already adds a header that says "Pumps" and you didn't change that. This seems extraneous...

lib/Ohmbrewer_RIMS.cpp
kyleoliveira: K

lib/Ohmbrewer_RIMS.cpp
```diff
@@ -253,25 +246,27 @@ int Ohmbrewer::RIMS::doDisplay(Ohmbrewer::Screen *screen) {
...
-    displayTunTemp(screen);
-
-    screen->printMargin(2);
-    screen->print("------  Tube  ------");
-    screen->printMargin(2);
+//    displayTunTemp(screen);
+//
+//    screen->printMargin(2);
+//    screen->print("----- RIMS Tube -----");
+//    screen->printMargin(2);
```
kyleoliveira: Clear out this commented-out stuff if you're not going to use it.

lib/Ohmbrewer_Onewire.cpp
azyth, line 157: above thing doesnt work i have tried it, and i may still need to print the car one at a time.

lib/Ohmbrewer_Pump.cpp
azyth, line 61: I will be eventually deleting that header, since it takes up alot of room.

lib/Ohmbrewer_Relay.cpp
kyleoliveira, line 251: ¯_(ツ)_/¯
I guess... if you're not paying attention...
However, that can only be true for single-pin Pumps.

lib/Ohmbrewer_Pump.cpp
kyleoliveira, line 61: If you're going to delete the header, then do it, and to be consistent you should do the same for at least heating elements and probably sensors too.
However, I wouldn't (or at least I'd leave it to another PR).
Think of it this way: how many pieces of equipment can we have other than sensors? 6 or less, right? RIMS + 1 other, 2 thermostats? With the header, we can display all of those things. Now, we might say that we can wire all the sensors to one power pin (in addition to the bus) and so we get a stupid high number of temperatures on the display, but I'd argue that's more of an argument for refactoring the temperature display section than anything else (possibly displaying 2 readings per line, for example). Ditto if for some reason we can squeeze more relay based stuff under one heading.
If you're trying to handle the use case where we have just a single pump and something else, then I'd suggest not using the displayPumps method at all. Either make a separate displayPump method that appends that text in front so it's clear you're only interested in a single pump, or do some special stuff, like we do with the recirculation pump in RIMS.

firmware/rhizome.ino
```diff
@@ -7,8 +7,9 @@
 #include "Ohmbrewer_Publisher.h"
 #include "Ohmbrewer_Thermostat.h"
 #include "Ohmbrewer_RIMS.h"
+#include "Ohmbrewer_Onewire.h"
 #include "onewire.h"
-// #include "Ohmbrewer_Publisher.h" // (*)
+#include <list>
```
kyleoliveira: You should include list down below the undef's where we include deque. Those undef's guard us against some Particle-specific shenanigans.
Again, the only reason we're not running into problems here is because list get's included earlier.

lib/Ohmbrewer_Onewire.cpp
kyleoliveira, line 157: Alrighty, then leave it.

I had a comment in here suggesting breaking the calls to screen->print into a separate method so you don't have to repeat it, but it seems to have disappeared. You should still do that - something like:

``` c++
void methodName(id, temp) {
screen->print(id);
screen->print(":");
screen->println(temp);
}
```

(approximately, of course)

lib/Ohmbrewer_RIMS.cpp
kyleoliveira: Added an issue for that ( #33 )

lib/Ohmbrewer_Relay.cpp
```diff
@@ -1,36 +1,93 @@
...
+ * Constructor
+ * @param id The Sprout ID to use for this piece of Equipment
+ * @param relayPins - controlPin always first in <list>
+ *  controlPin - The Control pin - Data/speed/power level Digital pin number X.
+ *  powerPin - The power pin - on/off line. Digital pin number X.
+ * @param stopTime The time at which the Equipment should shut off, assuming it isn't otherwise interrupted
+ * @param state Whether the Equipment is ON (or OFF). True => ON, False => OFF
+ * @param currentTask The unique identifier of the task that the Equipment believes it should be processing
+ */
+Ohmbrewer::Relay::Relay(int id, std::list<int>* relayPins, int stopTime,
```
kyleoliveira, line 70: This one and Relay(id, relayPins) do the same stuff (+ some extra here). Put that same stuff into a separate method ( initPins or some such), so there's only one place to update it if necessary.

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -176,11 +78,94 @@ Ohmbrewer::Thermostat::~Thermostat() {
 }
 
 /**
+ * logic for initializing the constructors
+ */
+void Ohmbrewer::Thermostat::initThermostat(int id, std::list<int>* thermPins){
```
kyleoliveira, line 85: Better, but this is still pretty hairy. I'd split it into initPID and initPins, and rename this "init". Then this method would pretty much be:

``` c++
initPID();
initPins(thermPins);
_targetTemp = new Temperature(-69);
```

firmware/rhizome.ino
```diff
@@ -76,20 +80,17 @@ void setup() {
 //    sprouts.push_back(new Ohmbrewer::HeatingElement( 2, new std::list<int>(1,5), 0, true, fakeTask ));
 
 // EX 4: A RIMS w/ a Pump to the Kettle
```
kyleoliveira, line 82: Don't forget to update the other examples (1-3) above so that we know we haven't broken anything... I tried using Thermostat here and I get a hard fault, despite commenting things out in the setup() and loop().

EDIT: Because I missed that you do the pushing of thermPins[1..2] right above the RIMS addition.

firmware/rhizome.ino
```diff
@@ -76,20 +80,17 @@ void setup() {
 //    sprouts.push_back(new Ohmbrewer::HeatingElement( 2, new std::list<int>(1,5), 0, true, fakeTask ));
```
kyleoliveira: I think Ex 3 ends up like:

``` c++
    std::list<int> elm1Pins (1, 3);
    elm1Pins.push_back(4);
    std::list<int> elm2Pins (1, 5);
    elm2Pins.push_back(6);

    sprouts.push_back(new Ohmbrewer::Pump( 1, 1, 0, true, fakeTask ));
    sprouts.push_back(new Ohmbrewer::Pump( 2, 2 ));
    sprouts.push_back(new Ohmbrewer::HeatingElement( 1, &elm1Pins ));
    sprouts.push_back(new Ohmbrewer::HeatingElement( 2, &elm2Pins, 0, true, fakeTask ));
```

right?

lib/Ohmbrewer_Temperature_Sensor.cpp
```diff
@@ -2,42 +2,30 @@
...
- * Sets the last time the temperature was read by the sensor
- * @param lastReadTime The last temperature reading time
+ * Constructor
+ * @param id The Sprout ID to use for this piece of Equipment
+ * @param busPin The Digital Pin that the temp probes are attached to. NOTE: for ds18b20 should always be D0
  */
-const int Ohmbrewer::TemperatureSensor::setLastReadTime(const int lastReadTime) {
-    _lastReadTime = lastReadTime;
-    return 0;
+Ohmbrewer::TemperatureSensor::TemperatureSensor(int id, int busPin) : Ohmbrewer::Equipment(id) {
```
kyleoliveira, line 11: Since we don't provide a Onewire constructor that picks a provided bus pin, we should drop the bus pin parameter to this constructor for now.

firmware/rhizome.ino
azyth: looks right to me, FYI you can do it with either 1 or 2 pins in the list

kyleoliveira: Bueno. Maybe make elm2Pins only one pin long when you add this in then.

lib/Ohmbrewer_Pump.cpp
kyleoliveira, line 61: Agreed over Hangouts that the display refactoring should go into a new Issue/PR.

lib/Ohmbrewer_Temperature_Sensor.cpp
kyleoliveira, line 11: Agreed over Hangouts that this should get spun off into another issue/PR. Created #34

lib/Ohmbrewer_Onewire.cpp
azyth, line 157: will do, this will be part of the display PR

lib/Ohmbrewer_Thermostat.h
```diff
@@ -215,6 +227,28 @@ namespace Ohmbrewer {
...
+            double aggKp=4;
+            double aggKi=0.2;
+            double aggKd=1;
+            double consKp=1;
+            double consKi=0.05;
+            double consKd=0.25;
+
+            //PID windowing variables
+            int windowSize = 5000;
+            unsigned long windowStartTime;
```
kyleoliveira, line 277: These PID-related variables should be static const. Otherwise they're associated with particular instances and can be modified during operation (which I don't think we want to be able to do at the moment).

kyleoliveira:
We're abandoning this branch, right?

PR #30, opened 2015-10-23
Dynamically adds/removes Equipment

kyleoliveira:
@azyth Take a look at this stuff and let me know what you think. It's not quite ready to go yet (there's some cleanup to do) - I'm mostly looking for double-checking I'm passing the pins in in the right order and so forth.

I'm also kicking around the idea of pulling the deque management stuff into a class of its own as I'm not a fan of it all just sitting in the .ino.

Finally, you'll note that while the add/remove functions should work, doing it this way seems to make it so the update functions for the Equipment objects doesn't register. I've got a question about that out on the Particle Community boards... If we have to define Particle.function calls within setup(), I might have to add a router function.

### DIFF
--- a/firmware/rhizome.ino
+++ b/firmware/rhizome.ino
@@ -9,13 +9,15 @@
 #include "Ohmbrewer_RIMS.h"
 #include "Ohmbrewer_Onewire.h"
 #include "onewire.h"
+#include "Ohmbrewer_Sprouts.h"
+// #include "Ohmbrewer_Publisher.h" // (*)
 
 // Kludge to allow us to use std::list - for now we have to undefine these macros.
 #undef min
 #undef max
 #undef swap
 #include <deque>
-//#include <list>
+#include <list>
 
 /* ========================================================================= */
 /*  Global Vars                                                              */
@@ -27,19 +29,19 @@
  */
 std::deque< Ohmbrewer::Equipment* > sprouts;
 
-/**
- * This is simply a map of which pins are currently accounted for. We'll want to
- * update it within the Sprout management functions.
- */
-int inUse[6] = {0,0,0,0,0,0};
-
 //initializer Pin list
-std::list<int> thermPins (1, 0);
+//std::list<int> thermPins (1, 0);
 
 /**
  * The touchscreen object. Handles the display for the Rhizome.
  */
 Ohmbrewer::Screen screen = Ohmbrewer::Screen(D6, D7, A6, &sprouts);
+
+/**
+ * The managed list of sprouts
+ * @todo Refactor how we handle Sprouts/SproutList so that the deque is encapsulated and Screen takes the list object
+ */
+Ohmbrewer::Sprouts sproutList = Ohmbrewer::Sprouts(&sprouts, &screen);
 
 unsigned long lastUpdate = millis();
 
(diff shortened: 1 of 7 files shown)


azyth:
So an issue with this is that alot is the constructors have changed again. Otherwise seems like a good plan. Will look more in depth later

kyleoliveira:
Yeah, I had a feeling that might be the case from what you've told me about what you've been working on. Let's figure out sometime this weekend or early this week where we can actually nail that down so that we can move forward.

firmware/rhizome.ino
```diff
@@ -98,22 +101,316 @@ void setup() {
...
+    for (itr; itr != sprouts.end(); itr++) {
+        if (((*itr)->getID() == id) && (type.equalsIgnoreCase((*itr)->getType()))) {
+            delete params;
+            return -2;
+        }
+    }
+
+    // Now, depending on the type we need to parse differently. Then we add the Equipment.
+    if(type.equalsIgnoreCase(Ohmbrewer::Pump::TYPE_NAME)) {
+        String powerPin   = String(strtok(NULL, ","));
```
azyth: this section will probably need to be gone over in more detail once the PID branch is merged since most/all pumps will currently be only power pin.

firmware/rhizome.ino
```diff
@@ -98,22 +101,316 @@ void setup() {
...
+        }
+        if(arePinsInUse(&pinTest)) {
+            delete params;
+            return -3;
+        }
+
+        sprouts.push_back(new Ohmbrewer::HeatingElement( id, powerPin.toInt(), controlPin.toInt() ));
+
+    } else if(type.equalsIgnoreCase(Ohmbrewer::Thermostat::TYPE_NAME)) {
+        int pins[3] = {-1, -1, -1};
```
azyth: hard coded pins being passed?

firmware/rhizome.ino
kyleoliveira: Yup, that's why I put it up for code review :smile:

azyth:
Hold til PID is ready

azyth:
you should be able to complete this now that the refactoring is done.

kyleoliveira:
@azyth This is ready for review again.

Should be ready for merge. As I mentioned above, this breaks the update functions, but I'm planning to handle that in a separate PR/Issue.

azyth:
This looks alot better, I think your conflicts are probably coming from screen now that display is merged in. So just make sure to test the display after you merge. 

So after you fix the conflicts if it flashes ok it seems like this will be good to go.

lib/Ohmbrewer_Screen.cpp
```diff
@@ -189,7 +211,9 @@ unsigned long Ohmbrewer::Screen::displayRelays() {
             ((Ohmbrewer::Relay*)(*itr))->display(this);
         }
     }
-    printMargin(2);
+    if(foundFirst) {
+        printMargin(2);
```
azyth, line 215: does it really need the second one here?

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -323,7 +331,7 @@ int Ohmbrewer::Thermostat::doDisplay(Ohmbrewer::Screen *screen) {
     // Print the section title
     screen->print("== Thermostat #");
     screen->print(getID());
-    screen->println(" ==");
+    screen->print("  ==");
```
azyth, line 334: does this work? I had to switch this to println becasue a print margin didnt automatically make a new line.

azyth:
the sprout stuff looks good, my comments on the display stuff are mostly cosmetic if they look good then they are fine. 

I have added a section of code to pid branch to remove the bug that would occur if you actually pass -1 as a pin number. so if this merges we will just need to be careful until that merges also. unless you want to add that here as well. (the initRelay() breakout and  contains the guard at the bottom now)

lib/Ohmbrewer_Screen.cpp
kyleoliveira, line 215: Was just making it consistent with the other ones. They all do this and just looking at Relay::doDisplay vs, for example, TemperatureSensor::doDisplay they end the same way breaks-wise (a println at the end).

lib/Ohmbrewer_Thermostat.cpp
kyleoliveira, line 334: It works for me, though now that you mention it, that does seem strange. 
The Screen::printMargin method does call println, so I'm wondering if the "right" way to do that would be to do a println here and remove the printMargin on line 337.

kyleoliveira:
I think it would be better to handle the -1 stuff outside this PR since it's really more of a separate bug and, unlike the issue with initThermostats, it doesn't kill this functionality, persay.

I'll take a look at the stuff you commented on and merge if the display works.

lib/Ohmbrewer_Thermostat.cpp
kyleoliveira, line 334: Looks like we do the same thing with RIMS, and both appear to look right, so I'm gonna leave it as-is.

Issue #39, opened 2015-11-02
Refactor display


azyth:
refactor the display for all classes to consolidate and make better UI

azyth:
currently on branch "disp"

PR #40, opened 2015-11-02
Display refactoring 

azyth:
changes the display around.  We should have  meeting that talks about our goal for the UI for this. since we have just been going back and forth but never actually came up with and agreed on a plan. 
#39


azyth:
add dispOther() to screen , list other (non mode) relays by type 

refactor doDisplay() in RIMS, THerm? other MODE level constructs. 

handle switching display in screen if other disp functions are called (menu, sel, dispProbeIDS...etc)

PR #43, opened 2015-11-07
refactors tempprobe constructor inputs

### LINKED ISSUE #34: TemperatureSensor constructors should take a Probe object (not pins)
We designed the TemperatureSensor so that it has an internal Probe object that takes care of any technical concerns revolving around reading the sensor such as whether you need one or more pins or how the hardware is identified (by pin or by bus ID). However, the TemperatureSensor constructors are currently set up to take a bus pin setting and are all hardwired to use the Onewire probe.

We should refactor all of the TemperatureSensor constructors so they take a pointer to an appropriately initialized Probe object instead of any sort of pin setup info. For example, this:

``` c++
// signature
TemperatureSensor(int id, int busPin);

// used like
ts = TemperatureSensor(1, D0);
```

should change to this

``` c++
// signature
TemperatureSensor(int id, Probe probe);

// used like
ts = TemperatureSensor(1, new Onewire());
```

Thoughts?

azyth:
resolves #34

### DIFF
--- a/lib/Ohmbrewer_RIMS.cpp
+++ b/lib/Ohmbrewer_RIMS.cpp
@@ -1,6 +1,7 @@
 #include "Ohmbrewer_RIMS.h"
 #include "Ohmbrewer_Screen.h"
 #include "Ohmbrewer_Publisher.h"
+#include "Ohmbrewer_Onewire.h"
 
 
 /**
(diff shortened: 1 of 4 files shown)


azyth:
@kyleoliveira short and sweet, unlike any PR you have ever seen from me before.

kyleoliveira:
Looks good to me! Go ahead and merge.

PR #44, opened 2015-11-07
Rims dowork

### LINKED ISSUE #28: Implement doWork for RIMS
Once we've got doWork workin' for the other Equipment, RIMS should be a piece of cake...

Note that this is pretty much dependent on #8 and #9 wrapping up.

azyth:
resolves #28  with two rims versions, I assume fancy rims will be the end result code but no sure if we want to have a simpler version or both?

### DIFF
--- a/lib/Ohmbrewer_RIMS.cpp
+++ b/lib/Ohmbrewer_RIMS.cpp
@@ -1,6 +1,7 @@
 #include "Ohmbrewer_RIMS.h"
 #include "Ohmbrewer_Screen.h"
 #include "Ohmbrewer_Publisher.h"
+#include "Ohmbrewer_Onewire.h"
 
 
 /**
@@ -80,7 +81,7 @@ Ohmbrewer::RIMS::~RIMS() {
 void Ohmbrewer::RIMS::initRIMS(int id, std::list<int>* thermPins, int pumpPin){
     int size = thermPins->size();
     if ( (size == 2) || (size == 3) ){
-        _safetySensor = new TemperatureSensor(id+2, thermPins->front());  //TODO need UIDS
+        _safetySensor = new TemperatureSensor(id+2, new Onewire());  //TODO need UIDS
         _tube = new Thermostat(id+3, thermPins);
     }else{
         //publish error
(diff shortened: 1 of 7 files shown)


lib/Ohmbrewer_RIMS.cpp
```diff
@@ -245,12 +246,58 @@ bool Ohmbrewer::RIMS::isOff() const {
...
+        // turn off therm
+        getTube()->setState(false);
+        getTube()->work(); //reset
+    }*/
+
+    //FANCY RIMS, turns the pump off to rest when tun temp is good and safety temp is good.
+    if (getState()) {
+        //IF RIMS ON
+        // make sure R. PUMP is ON if safety temp > tun temp +3(margin)
+        if (getSafetyTemp()->c() > (getTube()->getSensor()->getTemp()->c() + 3) ){
```
kyleoliveira: Why not use `getTunSensor()` instead of `getTube()->getSensor()` ?

lib/Ohmbrewer_RIMS.cpp
```diff
@@ -245,12 +246,58 @@ bool Ohmbrewer::RIMS::isOff() const {
...
+            if (!getRecirculator()->getState()){ getRecirculator()->setState(true); }// turn on pump
```
kyleoliveira: Break this if-statement into multiple lines for readability.

lib/Ohmbrewer_RIMS.cpp
```diff
@@ -245,12 +246,58 @@ bool Ohmbrewer::RIMS::isOff() const {
...
+            getRecirculator()->work();
+        }
+        //if tun temp is less than 2 degrees (margin) of target temp, then: RIMS
+        if (getTube()->getSensor()->getTemp()->c() < (getTube()->getTargetTemp()->c() - 2) ){
+            // safetySensor guard on Therm (if: safety sensor temp > safety setting, then: NO heat )
+            if ( getSafetyTemp()->c() > getSafetySensor()->getTemp()->c() ) {
+                if (!getTube()->getState()) { getTube()->setState(true); } // turn on therm
```
kyleoliveira: Break this if-statement into multiple lines for readability.

azyth:
NOTE: all equipment (electrical relay type) should have work() called on tehm after they have had their state turned off to make sure the pin outpins are turned off.

lib/Ohmbrewer_RIMS.cpp
azyth: because i forgot it existed

kyleoliveira:
Looks good to go.

Issue #42, opened 2015-11-03
Add Timers to control PID update intervals


azyth:
utilize the new software timers package to increase PID accuracy .  
https://docs.particle.io/reference/firmware/photon/#software-timers

kyleoliveira:
We should be able to either implement this or close the Issue if we don't want to attempt it. Take a look at the periodic update timer for an example of how to set up the appropriate pointer object.

@azyth what do you want to do?

azyth:
I am not sure, I think fro now we are ok as it is, but eventually we will want to integrate a different timing method.  is the periodic update timer part of the firmware package under the link above? or is it something separate. 
oh do you mean like #88 ?

kyleoliveira:
It's implemented as an instance of one of the provided Timer class, yes. It's just managed by the Rhizome class instance rather than a particular Equipment instance.

azyth:
I think that would be a worthwhile change. since we are currently not measuring how long our intervals are even, if i recall correctly.

kyleoliveira:
Ok, so to verify, we still want to add a Timer to each Thermostat (or each PID?) that will trigger the sampling or whatever for the PID to decide if it needs to change? 
Would we be able to implement that as a global Timer (within the Rhizome class) that triggers each PID in the Sprouts list? I realize it would be less accurate but I think there's a 10-instance limit on how many Timers you can instantiate. (Realistically, we only support ~3 Thermostats/PID's per Rhizome anyhow, admittedly. We'd probably want to put a check in there to make a hard limit of how many you can add to be safe.)

azyth:
I think for it to be worthwhile there needs to be a timer for each PID instance. where exactly it is kept dosnt matter, but each one should be able to keep track of its own cycle. The reason we didnt do this before was due to conflicting access to timers I think, as well as I think that the timers needed to be declared in the .ino file and I was having a hard time passing them or declaring them elsewhere.

kyleoliveira:
Ok, yeah, that's what I was thinking after typing all that. :grinning: 
As you can tell from #88, that's no longer an issue. I believe we'll want to keep the Timer in the PID class. We should also add a hard limit of 9 Sprouts with PID's within them (just Thermostat and RIMS for now).

azyth:
sounds good. want to adjust the issue? or create a new one with the specifics?

kyleoliveira:
Changed the title, but other than that I think we can keep this issue going as-is.

Issue #35, opened 2015-10-31
Move the aggressive/conservative PID profiles into a tiny class


kyleoliveira:
Right now we're hard-coding the default PID profiles as variables in the Thermostat class. It might be easier to work / cleaner with if we create a PID profile class along the lines of:

``` c++
public:
PIDProfile(kp, ki, kd) { /* set all the values */ };
int kP
int kI
int kD
```

Then we can refactor Thermostat with:

``` c++
// In the .h
const PIDProfile agg = PIDProfile(...);
const PIDProfile cons = PIDProfile(...);

// usage
_thermPID = new PID(&input, &output, &setPoint, cons.kP, cons.kI, cons.kD, PID::DIRECT);
_thermPID->SetTunings(agg); // if we want to add this capability to our fork of the PID lib
_thermPID->SetTunings(agg.kP, agg.kI, agg.kD); // if we don't
```

I realize this is somewhat low-gain, so that's why it's low priority. However, it's also fairly low effort and would be nice to have.

PR #45, opened 2015-11-09
adds timers 

azyth:
addresses #42

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,26 +5,29 @@ set(SOURCE_FILES
     ../firmware/rhizome.ino
     ../lib/Ohmbrewer_Equipment.h
     ../lib/Ohmbrewer_Equipment.cpp
-    ../lib/Ohmbrewer_Relay.h
-    ../lib/Ohmbrewer_Relay.cpp
     ../lib/Ohmbrewer_Heating_Element.h
     ../lib/Ohmbrewer_Heating_Element.cpp
     ../lib/Ohmbrewer_Onewire.cpp
     ../lib/Ohmbrewer_Onewire.h
     ../lib/Ohmbrewer_Probe.cpp
     ../lib/Ohmbrewer_Probe.h
+    ../lib/Ohmbrewer_Publisher.h
+    ../lib/Ohmbrewer_Publisher.cpp
     ../lib/Ohmbrewer_Pump.h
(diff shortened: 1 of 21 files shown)


azyth:
@kyleoliveira I am having trouble with the timers, i cannot find out a way to init them in the constructors and still have access to them elsewhere. They seem like they where designed primarily for the setup-loop structure.
-I cannot find much documentation specifying how to use them apart from the reference page. 
- I cannot figure out a way to assign a timer to a var so that you can call it from within a class object. 

any ideas?  the timer code is all in place but commented out if you want to see what i would like to be able to do. (initTherm & dowork() & thermostat.h _timer) primarily

kyleoliveira:
So, from what I'm gathering from the Particle docs, the Timer class is a callback model. Think of it kind of like tying a function to an interrupt. Every however many millis, it'll call that function.

As such,
1) There isn't much reason to access the Timer directly. It'll just do it's thang. I think you have the right idea with creating a `doPID` method.

2) Yeah, you're gonna want to create the Timer in the constructor. If we want to manage it (start/stop/rest), then you're probably going to want to set up an instance variable that's a Timer\* and use new to create it in the constructor (kind of like we do with the Temperature variable in TemperatureSensor). It looks like you were trying to do that with `_timer`, but I think the missing component is that it probably needs to be a pointer.

3) Another reason the pointer thing is probably the way to go because it'll allow you to call the dispose() method in the destructor, which will free up one of the 10 available timers if we dynamically remove the equipment. I think we have to do that, but I could be wrong (maybe it disposes itself when the object dies / pointer deletes?).

4) If you don't want to have to constantly start/stop the Timer when the state toggles, you can simply put an if-statement in the function the Timer calls and only do the innards if the state is ON. It would mean the Timer is wasting cycles and could affect the other Timers, but I don't know if that's overly expensive (we'd need to test a bit).

kyleoliveira:
Also, I probably wouldn't try to jam #35 into this. The timer stuff seems challenging enough without mixing that in.

azyth:
Do you think I should make this work in the same way you got the update Timer to work?

kyleoliveira:
Yeah, go for it. I'm still waiting on a useful response to my post on the Particle forum. If we don't figure anything out by the time you're ready to merge, make sure to add an Issue for refactoring it.

PR #46, opened 2015-11-11
Refactors the display to meet spec #39

azyth:
solves #39

### DIFF
--- a/firmware/rhizome.ino
+++ b/firmware/rhizome.ino
@@ -57,38 +57,39 @@ unsigned long lastUpdate = millis();
 void setup() {
 //    Serial.begin(9600); // Enable serial for debugging messages
     String fakeTask = "fake";
+    thermPins.push_back(2); //control pin (relay pin)
+    thermPins.push_back(3); //power pin (switch pin)
 
     ow_setPin(D0); //This should later be accomplished by equipment setup OR constructor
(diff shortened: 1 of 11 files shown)


azyth:
@kyleoliveira ready for a review. code compiles and flashes well for several configurations.

lib/Ohmbrewer_Heating_Element.cpp
```diff
@@ -1,4 +1,5 @@
 #include "Ohmbrewer_Heating_Element.h"
+#include "Ohmbrewer_Screen.h"
```
kyleoliveira: Any reason you added this given that you didn't change the rest of the file?

lib/Ohmbrewer_Pump.cpp
```diff
@@ -1,4 +1,6 @@
 #include "Ohmbrewer_Pump.h"
+#include "Ohmbrewer_Publisher.h"
+#include "Ohmbrewer_Screen.h"
```
kyleoliveira: Any reason you added these given that you didn't change the rest of the file?

kyleoliveira:
Looks good to me. I'm gonna say go with it, though you should try adding like 2 Thermostats and 6 temps and seeing what happens.

PR #50, opened 2015-11-14
Adds a router endpoint (/update) that provides a way to pass in Equipment updates.

### LINKED ISSUE #47: Add a router method for dynamically added Equipment
Particle currently does not support registering Particle Functions after setup() fires, so once we've swapped over to using dynamically added Equipment (see #30 ), we'll lose our update functions.

To fix this, we'll want to add a single Particle Function that acts as a router and takes the provided update string, parses it, finds the right Equipment in the Sprout List, and passes the rest of the update string on to that Equipment's update method.

Make this function available at `/update` and let the first two comma-delimited arguments be type and ID, respectively.

kyleoliveira:
Should resolve #47 .

Gets our old update() methods working again, but now you send the request to `/update` and append the TYPE_NAME for the class to the front of the arguments string.

Example:

``` shell
args=rims,1,69,ON,1447474616,ON,OFF,100,ON,ON
```

Means that for RIMS 1, set:

```
current task: 69 (will end up matching a Task ID in the web app)
RIMS 1 State: ON
Stop Time: 1447474616 (Should be sometime in the future, but we don't check that at the moment)
Safety Sensor State: ON
R. Pump State: OFF
Target Temp: 100 (Celsius)
Thermostat Element State: ON
Thermostat Sensor State: ON
```

@azyth Take a look and let me know if we're good to go.

### DIFF
--- a/lib/Ohmbrewer_Pump.cpp
+++ b/lib/Ohmbrewer_Pump.cpp
@@ -7,7 +7,7 @@
  */
 Ohmbrewer::Pump::Pump(int id, int pumpPin) : Ohmbrewer::Relay(id, pumpPin) {
 
-    registerUpdateFunction();
+//    registerUpdateFunction();
 }
 
 /**FIXME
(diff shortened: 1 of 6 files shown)


azyth:
looks good to me.

PR #51, opened 2015-11-14
Adds a simple PID profile class for managing groups of PID constants.

### LINKED ISSUE #35: Move the aggressive/conservative PID profiles into a tiny class

kyleoliveira:
Should resolve #35 .

@azyth What do you think?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,7 +42,8 @@ set(SOURCE_FILES
     ../lib/Ohmbrewer_Screen.h
     ../lib/Ohmbrewer_Screen.cpp
     ../lib/Ohmbrewer_Sprouts.h
-    ../lib/Ohmbrewer_Sprouts.cpp)
+    ../lib/Ohmbrewer_Sprouts.cpp
+    ../lib/Ohmbrewer_PID_Profile.h)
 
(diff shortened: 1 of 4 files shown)


azyth:
sure, this a bit simplier then what I was thinking you meant, so go ahead with this and I will modify what I have to fit.

Issue #37, opened 2015-11-02
Make sure Rhizome is publishing status changes


kyleoliveira:
Ohmbrewer's web app is going to rely on updates about status, so we need to be sure that when stuff changes in (at least) the Heating Elements, Temperature Sensors, and Pumps a status message is published.

Coordinate with @kyleoliveira on the expected interface.

kyleoliveira:
@Ohmbrewer/web-team  Here are the times I would expect an update to be sent out:

| When | Class | What to send |
| --- | --- | --- |
| Every 15 seconds | TemperatureSensor | Current Sensor reading |
| When Stop Time is reached (and the Equipment shuts off) | Everybody | Equipment state (maybe) |
| ~~When State changes~~ | ~~Relay/Pump/Thermostat~~ | ~~Current state~~ |
| When Target Temperature reached | Thermostat | ??? (Probably a simple notification on the channel, since we're already getting regular sensor updates) |
| When Max Temperature is reached and cooldown starts | RIMS | ??? (Mostly just needs to go to the right event channel) |
| When cooldown ends | RIMS | ??? (Mostly just needs to go to the right event channel) |

For all of these I'd want to include the Equipment ID and the Current Task. 

I don't think we need notifications about State change for RIMS since its subsystems will probably be sending that information out. I'm on the fence as to whether we want a notification about state change (other than Stop Time shutoff) from HeatingElement since we'll be modulating it so much with the Thermostat. I don't think it would hurt aside from taking up processing time, but I don't know if we gain enough to make it worthwhile either.

What do you think? Did I miss anything?

azyth:
as for state change it is possible that it can be updated and changed very rapidly in a short period of time. so i think omitting the low levels is fine.

kyleoliveira:
Now that you mention it, I think you're right about the state, but for a different reason. The only one that's going to toggle state rapidly is the Heating Element (I mean you could do the same thing with Relay, but we don't support that currently).

[...]

PR #53, opened 2015-11-15
Adds convenience functions to Publisher for fewer lines to do a publish().

### LINKED ISSUE #33: Quicker Publisher constructor
Currently, when you want to publish something, we do it like so:

``` c++
publish_map_t pMap;
Publisher pub = Publisher(new String("error_log"), &pMap);
 pMap[String("list_check_rims")] = String("improperly formed array - RIMS(int, int<list>, int)");
pub.publish();
```

It would be really handy if we could just say:

``` c++
Publisher pub = Publisher(new String("error_log"), 
                                          new String("list_check_rims"), 
                                          new String("improperly formed array - RIMS(int, int<list>, int)")
                                          );
pub.publish();
```

or

``` c++
publish_map_t pMap;
Publisher pub = Publisher(new String("error_log"), 
                                          &pMap,
                                          new String("list_check_rims"), 
                                          new String("improperly formed array - RIMS(int, int<list>, int)")
                                          );
// Add other stuff to pMap, if needed
pub.publish();
```

kyleoliveira:
Also adds a Timer-based function for reporting all the TemperatureSensor temperatures. Normally I don't mash this stuff together, but I needed a way to demonstrate the new functions worked and it kills two or three birds with one stone, so here we are.

Should resolve #33 and handle the first part of #37 .

Note that I had to do a janky delegation sort of thing to get the Timer to play nice with the class. From what I can tell, they didn't set up the Timer class the same way as Particle.function and there's no obvious way to pass in a reference to a particular instance of a class. I've tried every combination of &, *, ->, and lambda stuff I can think of, but no dice. [I've got a question about it on the Particle forums](https://community.particle.io/t/using-timer-in-a-class/17464), but I figure this will at least get us off the ground.

So the best I could do is instantiate a pointer to a Timer instance and pass it to the Sprouts object to manage (instead of instantiating it within Sprouts' constructor). When I instantiate the pointer, I provide it with a reference to a global function (like in the Particle docs) and that global function does the delegation. It's not pretty, I don't like it, but it works.

@azyth Good to go?

### DIFF
--- a/lib/Ohmbrewer_Sprouts.cpp
+++ b/lib/Ohmbrewer_Sprouts.cpp
@@ -14,12 +14,13 @@
  * @param sprouts A pointer to a pre-constructed deque object. Sprouts takes over responsibility of memory management for it.
  * @param screen A pointer to the Screen object. Sprouts does not handle its memory management.
  */
-Ohmbrewer::Sprouts::Sprouts( std::deque< Ohmbrewer::Equipment* > *sprouts, Ohmbrewer::Screen *screen) {
+Ohmbrewer::Sprouts::Sprouts( std::deque< Ohmbrewer::Equipment* > *sprouts, Ohmbrewer::Screen *screen, Timer *put) {
     _sprouts = sprouts;
     _screen = screen;
-    Spark.function("add", &Sprouts::addSprout, this);
-    Spark.function("update", &Sprouts::updateSprout, this);
-    Spark.function("remove", &Sprouts::removeSprout, this);
+    _periodicUpdateTimer = put;
+    Particle.function("add", &Sprouts::addSprout, this);
+    Particle.function("update", &Sprouts::updateSprout, this);
+    Particle.function("remove", &Sprouts::removeSprout, this);
 }
 
 /**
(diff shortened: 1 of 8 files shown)


lib/Ohmbrewer_Sprouts.cpp
```diff
@@ -101,6 +103,11 @@ int Ohmbrewer::Sprouts::addSprout(String argsStr) {
         return errorCode;
     }
 
+    if(_sprouts->size() == 1) {
+        // We should only have to start the update timer if this is the first Sprout
```
azyth, line 107: wont this only start the timer if there in only 1 thing in sprouts?

azyth:
looks good apart from the janky timer thing, but if it works id say go with it for now but add an issue so that we remember to fix it later once we can figure out how.

lib/Ohmbrewer_Sprouts.cpp
kyleoliveira, line 107: Correct. If size is >1, it's already running periodically (check the periodic update method - I restart the timer at the end) and if size == 0, I stop the timer. The alternatives would have been to start the Timer once in the constructor, or sometime before I add it to the Sprouts object.

kyleoliveira:
Created an Issue to keep track of the need to refactor.

Issue #52, opened 2015-11-14
Make sure Rhizome is using Stop Time


kyleoliveira:
Reviewing stuff for #37 and it seems that we're not doing anything with the supplied Stop Time variable. This member is meant to provide a fail-safe where, if no other command is given, the Equipment will automatically shut off (state = OFF) at this time. It's mostly meant to guard against network issues, and should not be the only fail safe since someone could conceivably set it to a time that's much too far in the future to be useful.

The simplest fix will probably be a matter of adding a check and some handling within Equipment::work(). Beyond that, we also probably need some handling in Thermostat and RIMS that passes the current Stop Time and Current Task information down into their subsystems.

UPDATE: This work should also include publishing an update to the cloud when the Equipment transitions to OFF.

PR #54, opened 2015-11-15
Adds status publishing when the RIMS safety sensor kicks in and when the Thermostat reaches its target temp.

kyleoliveira:
Hold until #53 is merged. Should resolve what remains of #37 except the Stop Time-related status messages. I think we should handle those when we deal with #52 .

@azyth What do you think? (You may need to click on the actual commit to see what applies to this PR specifically...)


azyth:
looks good.

PR #56, opened 2015-11-20
pid bug fixes and cmakelist cleanup

azyth:
@kyleoliveira this has the timer stuff disabled for now, but it resolves some issues with the thermostats to get them working, as well as alphabetizes the cmake list file.  also cleans up some of the logic in rims dowork().

### DIFF
--- a/lib/Ohmbrewer_Relay.cpp
+++ b/lib/Ohmbrewer_Relay.cpp
@@ -12,6 +12,7 @@
 Ohmbrewer::Relay::Relay(int id, int powerPin) : Ohmbrewer::Equipment(id) {
 
     _powerPin = powerPin;
+    pinMode(powerPin, OUTPUT);
     _controlPin = -1;
 }
 
@@ -26,6 +27,7 @@ Ohmbrewer::Relay::Relay(int id, int powerPin) : Ohmbrewer::Equipment(id) {
 Ohmbrewer::Relay::Relay(int id, int powerPin, int stopTime,
                         bool state, String currentTask) : Ohmbrewer::Equipment(id, stopTime, state, currentTask) {
     _powerPin = powerPin;
+    pinMode(powerPin, OUTPUT);
     _controlPin = -1;
 }
 
@@ -37,24 +39,7 @@ Ohmbrewer::Relay::Relay(int id, int powerPin, int stopTime,
  *  powerPin - The power pin - on/off line. Digital pin number X.
  */
 Ohmbrewer::Relay::Relay(int id, std::list<int>* relayPins) : Ohmbrewer::Equipment(id) {
-    int size = relayPins->size();
-    int powerPin = -1;
-    int controlPin = -1;
-    if  (size == 2){ // switch and relay
-        controlPin = relayPins->front();
-        powerPin = relayPins->back();
-    }else if (size == 1){ //relay only
-        controlPin = relayPins->front();
-    }else {//publish error
-        Ohmbrewer::Publisher::publish_map_t pMap;
-        Ohmbrewer::Publisher* pub = new Ohmbrewer::Publisher(new String("error_log"), &pMap);
-        pMap[String("list_check_relay")] = String("improperly formed input - Relay(int, int<list>)");
-        pub->publish();
-        delete pub;
-    }
-    //will be set to -1 for error checking if not enabled
-    _powerPin = powerPin;
-    _controlPin = controlPin;
+    initRelay(relayPins);
 }
 
 /**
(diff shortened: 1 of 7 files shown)


lib/Ohmbrewer_Relay.cpp
```diff
@@ -4,7 +4,7 @@
 
 
 /**
- * Constructor
+ * Constructor - PUMP
```
kyleoliveira: I don't think we need a comment about the Pump in Relay here. Kinda unnecessary.

lib/Ohmbrewer_Relay.cpp
```diff
@@ -110,6 +78,38 @@ Ohmbrewer::Relay::~Relay() {
...
+    int controlPin = -1;
+    if  (size == 2){ // switch and relay
+        controlPin = relayPins->front();
+        powerPin = relayPins->back();
+    }else if (size == 1){ //relay only
+        controlPin = relayPins->front();
+    }else {//publish error
+        Ohmbrewer::Publisher::publish_map_t pMap;
+        Ohmbrewer::Publisher* pub = new Ohmbrewer::Publisher(new String("error_log"), &pMap);
+        pMap[String("list_check_relay")] = String("improperly formed input - Relay(int, int<list>)");
```
kyleoliveira: Since you're moving all this stuff into a cleaner function anyway, you should probably use the simpler Publisher constructor so you don't have to have so many lines.
You also shouldn't have to create it as a pointer (or, subsequently, mess with delete).

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -258,6 +258,37 @@ bool Ohmbrewer::Thermostat::isOff() const {
  * Performs the Thermostat's current task. Expect to use this during loop().
  * This function is called by work().
  *
+ *
+ * @returns The time taken to run the method
+ */
+int Ohmbrewer::Thermostat::doWork() {
+    unsigned long start = micros();
+
+    if (getState()){
```
kyleoliveira, line 267: You can also use `isOn()` if you want to be all semantic.

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -258,6 +258,37 @@ bool Ohmbrewer::Thermostat::isOff() const {
...
+        doPID();
+    }else{
+        //Shut down procedure
+        //if thermostat is turned off then turn off element too.
+        getElement()->setState(false);
+        getElement()->work();//reset element
+        //stop Timer
+//        _timer->stop();
+        //if PID is off then update temp sensor
+        getSensor()->work();
```
kyleoliveira: You should shut the sensor off too. Otherwise it's going to keep sending temperature updates.

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -292,24 +322,24 @@ int Ohmbrewer::Thermostat::doWork() {
...
     }
     //TURN OFF
-    if (gap == 0) {//once reached target temp
-        _heatingElm->setState(false); //turn off element
-        if (_heatingElm->getPowerPin() != -1) { // if powerPin enabled
-            digitalWrite(_heatingElm->getPowerPin(), LOW); //turn it off too
+    if (gap == 0 || getTargetTemp()->c() <= getSensor()->getTemp()->c() ) {//once reached target temp
+        getElement()->setState(false); //turn off element
+        if (getElement()->getPowerPin() != -1) { // if powerPin enabled
+            digitalWrite(getElement()->getPowerPin(), LOW); //turn it off too
```
kyleoliveira: I don't think you want to mess with the digitalWrite for the power pin here. The Element's doWork function should handle it's pins appropriately depending on its given state.

lib/Ohmbrewer_Thermostat.cpp
```diff
@@ -292,24 +322,24 @@ int Ohmbrewer::Thermostat::doWork() {
...
     if (getState() && gap!=0) {//if we want to turn on the element (thermostat is ON)
         //TURN ON state and powerPin
-        if (!(_heatingElm->getState())) {//if heating element is off
-            _heatingElm->setState(true);//turn it on
-            if (_heatingElm->getPowerPin() != -1) { // if powerPin enabled
-                digitalWrite(_heatingElm->getPowerPin(), HIGH); //turn it on (only once each time you switch state)
+        if (!(getElement()->getState())) {//if heating element is off
+            getElement()->setState(true);//turn it on
+            if (getElement()->getPowerPin() != -1) { // if powerPin enabled
+                digitalWrite(getElement()->getPowerPin(), HIGH); //turn it on (only once each time you switch state)
```
kyleoliveira, line 328: It would be good to add a pair of methods to HeatingElement that handle flipping this pin for the modulation, if only to shorten these lines up a bit. Something like

``` c++
getElement()->modulateHigh();
getElement()->modulateLow();
```

Then Thermostat wouldn't need to know which pin is the pin to modulate - it would help with encapsulation and all that.

lib/Ohmbrewer_Thermostat.cpp
kyleoliveira: If you're setting the state to OFF, it should already be shutting off the Power pin.

azyth: i should probably replace lines 341 and 342 with getElement-> work()

lib/Ohmbrewer_Relay.cpp
azyth: I wanted to make sure people knew what this specific  constructor was for.  suppose it isnt necessary

PR #58, opened 2015-11-21
Adds documentation about the Rhizome's REST API.

kyleoliveira:
@azyth Is this sufficient? Questions?


kyleoliveira:
May best be viewed on the branch instead of via the diff screen, just fyi. It's all Markdown-y, so GitHub should interpret the tables and stuff better if it isn't in a code diff.

PR #61, opened 2015-11-22
multiple onewire Probes

### LINKED ISSUE #49: Multiple onewire probes
almost forgot about this one. we can handle it currently but we just need a way to identify them to the user.

azyth:
resolves #49


azyth:
so one option I thought about is making a master index list (say in sprouts) of existing onewire probes installed, and as part of the constructor you have to plug them in and add one at a time, and it consults the masterlist, finds which probe is new, adds that one as the new probe, then adds it to the master list. 

currently trying to accomplish this via casting the ID to an int and using casting to compare connected probes. still a few issues since i am not getting a proper reading.

lib/Ohmbrewer_Onewire.cpp
```diff
@@ -108,14 +123,17 @@ double Ohmbrewer::Onewire::getReading(){
...
     screen->resetTextColor();
+    screen->println("-------------------");
     // Print the section title
-    screen->print("PROBE ID# : TEMP");
+    screen->println("PROBE ID# : TEMP");
 //    screen->println("");
     screen->printMargin(2);
 
-
+    char msg[100];
```
kyleoliveira, line 133: Given that we only print 20 characters per line, this seems like a little much...

lib/Ohmbrewer_Onewire.cpp
```diff
@@ -124,7 +142,7 @@ void Ohmbrewer::Onewire::displayProbeIds(Ohmbrewer::Screen *screen){
     //If your code has other tasks, you can store the timestamp instead and return when a second has passed.
     delay(1000);
     uint8_t numSensors = ow_search_sensors(10, sensors);
-    screen->displayHeader();
+
     for (uint8_t i=0; i<numSensors; i++){
         //current probe ID
         char probeId[8] = {
```
kyleoliveira, line 145: I think we technically would want to use `byte` instead of `char` here, since we're working with unsigned HEX values. Obviously works either way though.

lib/Ohmbrewer_Sprouts.cpp
```diff
@@ -9,6 +9,9 @@
 #include "Ohmbrewer_RIMS.h"
 #include "Ohmbrewer_Onewire.h"
 
+
+//TODO onewire index refactor ??
+
```
kyleoliveira: TODO:
1) Refactor `Ohmbrewer::Sprouts::parseTemperatureSensorPins` so that instead of `busPin` it has `probeIndex` (just a rename, really)
2) Refactor `Ohmbrewer::Sprouts::addTemperatureSensor` so that instead of `busPin` it has `probeIndex` (just a rename, really). Then, change the TemperatureSensor constructor it calls is `TemperatureSensor(int ID)`.
3) Refactor the `TemperatureSensor(int ID)` constructor to automatically use the ID as the `probeIndex` in the `Onewire(int probeIndex)` constructor for its probe.
4) Refactor `Ohmbrewer::Sprouts::parseThermostatPins` so that instead of `sensorBus` it has `probeIndex` (just a rename, really)
5) Refactor `Ohmbrewer::Sprouts::parseRIMSPins` so that it expects the Safety Sensor's probeIndex in between the Thermostat pins and the Pump pins. You can pretty much lift the sensor code block from `Ohmbrewer::Sprouts::parseThermostatPins` and rename stuff and you'll probably be in business there. Once you've added that block, you'll need to add another parameter to the function to return the parsed value. The format of it will be similar to how the pump pin gets returned.
6) Refactor `Ohmbrewer::Sprouts::addRIMS` to handle the safety sensor `probeIndex` being returned from `Ohmbrewer::Sprouts::parseRIMSPins`. You'll need to take the returned value and plug it into the right place in the RIMS constructor, naturally.
7) Refactor the RIMS and Thermostat constructors so that they can handle the `probeIndex`'es appropriately. It looks like you've already started on this (or at least noted it).
8) Update the README/docs to include the new argument in the argument strings for /add.

I think if you do all of those things, we'll be looking pretty good. That should get /add working with the IDs/Indices and the other Particle.functions call everything by Type+ID pairs so we shouldn't need to do anything to fix them.

azyth:
take a look at this, if seems to be ok, I will do soem testing with it later, but let me know if you see anything that looks wrong or missing. (mainly want your input on sprouts as you wrote the original)

lib/Ohmbrewer_Sprouts.cpp
```diff
@@ -315,15 +318,18 @@ bool Ohmbrewer::Sprouts::arePinsInUse(std::list<int>* newPins) {
  * Parses a given string of characters into the pins for a Temperature Sensor
  * @param params The buffer to use for strtok'ing. This method will not delete the buffer!
  * @param busPin The bus pin
+ * @param index The onewire sensor index (-1 if unused / non onewire)
  * @return Error or success code, according to the requirements specified by addSprout
  */
-int Ohmbrewer::Sprouts::parseTemperatureSensorPins(char* params, int &busPin) {
-    String pin = String(strtok(NULL, ","));
+int Ohmbrewer::Sprouts::parseTemperatureSensorPins(char* params, int &busPin, int &index) {
```
kyleoliveira: Refactor the name of the method if you'd like, but I think we shouldn't take the bus pin in here. We only support OneWire sensors at the moment, and we only support using a single pin for the bus. There's no real reason to pass it in every time. Instead, if we want to support changing the OneWire bus pin, we can add a new Particle.function for that. It's not like our OneWire library really supports running more than one bus at a time (if you want that) anyway - the function to set the pin is global.

lib/Ohmbrewer_Sprouts.cpp
```diff
@@ -493,6 +504,17 @@ int Ohmbrewer::Sprouts::parseRIMSPins(char* params, std::list<int> &thermPins, i
...
     errorCode = parsePumpPins(params, pumpPin);
+    if(errorCode != 0) {
+        return errorCode;
+    }
+
+    String safe = String(strtok(NULL, ","));
+    if (safe == NULL){
+        return -1;
+    }//TODO add other error codes?
+    safetyIndex = safe.toInt();
```
kyleoliveira: If we don't provide the sensor bus pin in the argument string, this part can simply use the parseTemperatureSensorPins function.

azyth:
@kyleoliveira all ready to go if you like the sprouts changes.

lib/Ohmbrewer_Sprouts.cpp
```diff
@@ -409,8 +410,10 @@ int Ohmbrewer::Sprouts::parseHeatingElementPins(char* params, std::list<int> &el
 
     // Verify that the pins are not in use
     elementPins.push_back(controlPin.toInt());
-    if(powerPin != NULL && powerPin.toInt() > -1) {
+    if(powerPin != NULL ) {
```
kyleoliveira: What if it's a -2?

If you want to make this more correct/protective, you would probably want to restrict the inputs from -1 to 6, inclusive and return an error code indicating bad pin values if outside that range.

lib/Ohmbrewer_Sprouts.cpp
```diff
@@ -409,8 +410,10 @@ int Ohmbrewer::Sprouts::parseHeatingElementPins(char* params, std::list<int> &el
...
         elementPins.push_back(powerPin.toInt());
+    }else{
+        elementPins.push_back(-1);//powerPin not specified
```
kyleoliveira: The constructor handles this based on the size of the pin list, so this shouldn't be necessary.

lib/Ohmbrewer_Onewire.h
```diff
@@ -33,16 +33,27 @@ namespace Ohmbrewer {
...
          */
-        void setProbeId(uint8_t (&probeId)[8]);
+        void setProbeIndex(int index);
 
         /**
-         * @returns _probeID Unique probe ID
+         * @returns _probeIndex Unique (onewire) probe index ID
          */
-        uint8_t* getProbeId();
+        int getProbeId();
```
kyleoliveira: Should probably rename this to getProbeIndex

lib/Ohmbrewer_Onewire.cpp
azyth, line 145: it works

lib/Ohmbrewer_Onewire.cpp
azyth, line 133: it overruns the line anyways

lib/Ohmbrewer_Sprouts.cpp
azyth: i had origionally done this wehn you wanted to change it so -1s where passed along but it isnt needed so im fixing it.

kyleoliveira:
Good to go once you squash it down, methinks...

Issue #71, opened 2016-01-06
Build out basic menu feature


azyth:
implement the basic interactive menu for the display, 
- button press functionality
- basic menu
- navigation of menu

azyth:
@nischalvasant How is the pull request coming for this?

chillinginchaos:
I'm working on a state diagram for the menu... I'm trying to figure out the best way to keep track of "where I am" in the menu, and also how to go "back" to the screen I was on before I entered the menu.

kyleoliveira:
Resolved by #73 , though we need another issue for calibration. We should open additional issues for other screens/menus as needed.

Issue #68, opened 2015-11-29
boot if no WIFI available


azyth:
currently units do not boot if unable to connect to wifi. 

resource:
https://docs.particle.io/reference/firmware/photon/#wifi-on-

azyth:
add a check on doperiodic updates to check for connection to wifi. otherwise it might break

kyleoliveira:
Resolved by #73

PR #73, opened 2016-02-13
Adds touch screen functions. Detects when +,-,Menu,Sel buttons are pr…

chillinginchaos:
…essed.

Rhizome now starts without connecting to WiFi.
For demonstration purposes, It will only connect to WiFi when + button is pressed
A later update will include amenu setting to set and remember whether it should connec to WiFi at boot.

### DIFF
--- a/firmware/rhizome.ino
+++ b/firmware/rhizome.ino
@@ -18,6 +18,7 @@
 #include "Ohmbrewer_RIMS.h"
 #include "Ohmbrewer_Onewire.h"
 #include "Ohmbrewer_Sprouts.h"
+#include "Ohmbrewer_Runtime_Settings.h"
 //external libraries
 #include "onewire.h"
 
@@ -36,8 +37,11 @@
 // to a single piece of Equipment.
 std::deque< Ohmbrewer::Equipment* > sprouts;
 
+// Various settings used during the run
+Ohmbrewer::RuntimeSettings settings = Ohmbrewer::RuntimeSettings();
+
 // The touchscreen object. Handles the display for the Rhizome.
-Ohmbrewer::Screen screen = Ohmbrewer::Screen(D6, D7, A6, &sprouts);
+Ohmbrewer::Screen screen = Ohmbrewer::Screen(D6, D7, A6, &sprouts, &settings);
 
 // A timer for doing things every 15 seconds. Used by the sproutList below.
 Timer periodicUpdateTimer = Timer(15000, doPeriodicUpdates);
@@ -64,9 +68,9 @@ void setup() {
 	screen.initScreen();
 	
 	//Turn on for debugging
-	//Serial.begin(9600);
+	Serial.begin(9600);
 	
-	delay(1000);
+//	delay(1000);
 }
 
 /**
@@ -90,13 +94,13 @@ void loop() {
  */
 void doPeriodicUpdates() {
 	//Do not attempt to publish updates if disconnected from the cloud
-		if(!Particle.connected()){
-				//TODO: Define WiFi Address location properly
-				if(EEPROM.read(1) == 0x01){
-				//WiFi is not connected and should be - attempt to connect
-				Particle.connect();
-		}
+    if(!Particle.connected()){
+        // TODO: Define WiFi Address location properly
+        if(EEPROM.read(1) == Ohmbrewer::RuntimeSettings::EEPROM_WIFI_STATUS_OFF){
+            //WiFi is not connected and should be - attempt to connect
+            Particle.connect();
+        }
 	} else {
-		sproutList.publishPeriodicUpdates();
-	}
+        sproutList.publishPeriodicUpdates();
+    }
 }
--- a/firmware/rhizome.ino
+++ b/firmware/rhizome.ino
@@ -29,7 +29,7 @@
 #include <list>
 
 /* ========================================================================= */
-/*  Global Vars                                                              */
+/*	Global Vars																 */
 /* ========================================================================= */
 
 // Each Sprout is a logical collection of physical pins/relays that are connected
@@ -48,37 +48,55 @@ Ohmbrewer::Sprouts sproutList = Ohmbrewer::Sprouts(&sprouts, &screen, &periodicU
 unsigned long lastUpdate = millis();
 
 /* ========================================================================= */
-/*  Main Functions   (setup, loop)                                           */
+/*	Main Functions	 (setup, loop)											 */
 /* ========================================================================= */
 /**
  * Does any preliminary setup and initializations before the Rhizome starts the operation loop.
  */
+// Setting the photon to semi automatic mode, which means it 
+// does not connect to WiFi until Particle.connect() is called.
+SYSTEM_MODE(SEMI_AUTOMATIC);
 void setup() {
-    //initialize the Dallas Onewire bus pin - Digital 0
-    ow_setPin(D0); //This should later be accomplished by equipment setup OR constructor
+	//initialize the Dallas Onewire bus pin - Digital 0
+	ow_setPin(D0); //This should later be accomplished by equipment setup OR constructor
 
-    // Turn on the display
-    screen.initScreen();
+	// Turn on the display
+	screen.initScreen();
+	
+	//Turn on for debugging
+	//Serial.begin(9600);
+	
+	delay(1000);
 }
 
 /**
  * The bulk of the program. Runs repeatedly until the Rhizome is powered off.
  */
 void loop() {
-    //call work on all installed equipment in the sprouts list
-    sproutList.work();
 
-    //refresh the display
-    screen.refreshDisplay();
+	//call work on all installed equipment in the sprouts list
+	sproutList.work();
+	//check for button press and refresh the screen
+	screen.captureButtonPress();
+	screen.refreshDisplay();
 }
 
 /* ========================================================================= */
-/*  Other Global Functions                                                   */
+/*	Other Global Functions													 */
 /* ========================================================================= */
 
 /**
  * Delegation function that allows us to call the managed method for publishing periodic updates.
  */
 void doPeriodicUpdates() {
-    sproutList.publishPeriodicUpdates();
+	//Do not attempt to publish updates if disconnected from the cloud
+		if(!Particle.connected()){
+				//TODO: Define WiFi Address location properly
+				if(EEPROM.read(1) == 0x01){
+				//WiFi is not connected and should be - attempt to connect
+				Particle.connect();
+		}
+	} else {
+		sproutList.publishPeriodicUpdates();
+	}
 }
--- a/lib/Ohmbrewer_Menu.cpp
+++ b/lib/Ohmbrewer_Menu.cpp
@@ -1,5 +1,109 @@
 #include "Ohmbrewer_Menu.h"
 #include "Ohmbrewer_Screen.h"
+#include "Ohmbrewer_Runtime_Settings.h"
 
+/**
+ * Constructors
+ * @param screen A pointer to the Screen object.
+ * @param settings A pointer to the RuntimeSettings object.
+ */
+Ohmbrewer::Menu::Menu(Ohmbrewer::Screen *screen, Ohmbrewer::RuntimeSettings *settings) {
+    _screen = screen;
+    _settings = settings;
+}
 
+/**
+ * Destructor
+ */
+Ohmbrewer::Menu::~Menu() {
+    // Do nothing
+}
 
+/**
+ * Draws the menu to the Rhizome's display.
+ */
+void Ohmbrewer::Menu::displayMenu() {
+    // Disabled by default
+}
+
+/**
+ * Takes action when the menu button is pressed in the Menu Screen
+ * The precise action will be determined by the individual menu screens
+ */
+void Ohmbrewer::Menu::menuPressed() {
+    // Disabled by default
+}
+
+/**
+ * Takes action when the select button is pressed in the Menu Screen
+ * The precise action will be determined by the individual menu screens
+ */
+void Ohmbrewer::Menu::selectPressed() {
+    // Disabled by default
+}
+
+/**
+ * Takes action when the plus button is pressed in the WiFi Menu Screen
+ * @returns The time taken to run the method
+ */
+void Ohmbrewer::Menu::plusPressed(){
+    if(_selectedOption < (options.size() - 1)){
+        _selectedOption++;
+    } else {
+        _selectedOption = 0;
+    }
+}
+
+/**
+* Takes action when the minus button is pressed in the WiFi Menu Screen
+* @returns The time taken to run the method
+*/
+void Ohmbrewer::Menu::minusPressed(){
+    if(_selectedOption > 0){
+        _selectedOption--;
+    } else {
+        _selectedOption = (options.size() - 1);
+    }
+}
+
+/**
+ * Transitions to the next level up (parent) in the menu tree
+ * @returns The time taken to run the method
+ */
+void Ohmbrewer::Menu::goUp() {
+    _screen->setCurrentMenu(_parent);
+}
+
+/**
+ * Transitions to the next level down via a specified child in the menu tree
+ * @param child The index of the child to transition to within the menu's descendants list
+ * @returns The time taken to run the method
+ */
+void Ohmbrewer::Menu::goDownTo(const int child) {
+    _screen->setCurrentMenu(_children[child]);
+}
+
+/**
+ * Sets the parent of this menu
+ * @param parent The parent to set
+ */
+void Ohmbrewer::Menu::setParent(Menu* parent) {
+    _parent = parent;
+}
+
+/**
+ * Adds a child menu below this menu
+ * @param child The menu to add
+ */
+void Ohmbrewer::Menu::addChild(Menu* child) {
+    child->setParent(this);
+    _children.push_back(child);
+}
+
+/**
+ * True if this is the top-most menu in the tree
+ * @returns True if this is the Home menu, False otherwise
+ */
+bool Ohmbrewer::Menu::isHome() {
+    return _parent == nullptr; // Could just return _parent, but let's be explicit
+}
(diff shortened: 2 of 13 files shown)


chillinginchaos:
Addresses #68 (boots without WiFi) and part of #71 (Touch screen functionality)

lib/Ohmbrewer_Screen.h
```diff
@@ -1,13 +1,28 @@
...
+#define XM A7
+#define YM A2
+
+// This is calibration data for the raw touch data to the screen coordinates
+#define TS_MINX 500
+#define TS_MINY 300
+#define TS_MAXX 3650
+#define TS_MAXY 3650
+#define MINPRESSURE 50
+#define MAXPRESSURE 4000
```
kyleoliveira: Rather than using #define, make these static const variables. Compare against what was done for BUTTONSIZE, for example.

lib/Ohmbrewer_Screen.cpp
```diff
@@ -360,3 +362,97 @@ unsigned long Ohmbrewer::Screen::displayStatusUpdate(char *statusUpdate) {
...
+        }
+        
+        // Barf it onto the display...
+        statusUpdate.toCharArray(status, 40);
+        displayStatusUpdate(status);
+    }
+  	
+    Serial.print("X = "); Serial.print(p.x);
+    Serial.print("\tY = "); Serial.print(p.y);
+    Serial.print("\tPressure = "); Serial.println(p.z);
```
kyleoliveira: Comment out this debugging information if you want to include it in the commit.

lib/Ohmbrewer_Screen.cpp
```diff
@@ -360,3 +362,97 @@ unsigned long Ohmbrewer::Screen::displayStatusUpdate(char *statusUpdate) {
...
+unsigned long Ohmbrewer::Screen::captureButtonPress() {
+    unsigned long start = micros();
+    char printx [10];
+	char printy [10];
+	char status [40];
+	
+	// a point object holds x y and z coordinates
+  	TSPoint p = ts.getPoint();
+  	//According to Particle forums, the read below is necessary to get touch to work on Photon
+  	ts.readTouchY();
```
kyleoliveira: Do we only need to do this once after boot-up, or once every single time we're scanning for a touch? If it's the former, maybe this can be moved into the Screen constructor?

azyth: I think it must be done every loop , since we are capturing button presses continuously.

azyth:
@nischalvasant do we have any progress on this?

chillinginchaos:
I'm almost ready to do a new PR with a proper screen to modify and "remember" the WiFi selection... I'm hoping to get that out tonight.

azyth:
So this is looking good, I have a question on the proposed expansion scheme.
will you create a new class for each sub menu screen and link through them for navigation? thus allowing the functionality to change for the button press depending on which menu item you are in?  You created the file menu.c but the constructor creates ::wifimenu::

chillinginchaos:
Yes. Basically it will be a tree of Menu objects, and the screen will keep
track of which Menu object is currently selected and call the same
function, but executing whatever it is that Menu screen is supposed to do.
So like currentMenuScreen->plusPressed();

Nischal Vasant

firmware/rhizome.ino
```diff
@@ -53,22 +53,32 @@ unsigned long lastUpdate = millis();
...
  * The bulk of the program. Runs repeatedly until the Rhizome is powered off.
  */
 void loop() {
+
     //call work on all installed equipment in the sprouts list
     sproutList.work();
+ 
 
-    //refresh the display
+	screen.captureButtonPress();
```
kyleoliveira, line 84: I haven't downloaded this yet, so I'm not sure if you're using tabs or what, but make sure the indentation matches up for style. Pretty sure we're doing 4 space indents in this project judging by the other lines shown in the PR (though I could have sworn we used 2 spaces... hmm.)

firmware/rhizome.ino
```diff
@@ -80,5 +90,13 @@ void loop() {
  * Delegation function that allows us to call the managed method for publishing periodic updates.
  */
 void doPeriodicUpdates() {
-    sproutList.publishPeriodicUpdates();
+	//Do not attempt to publish updates if disconnected from the cloud
+    if(Particle.connected() == false){
+		if(EEPROM.read(1) == 0x01){
```
kyleoliveira: Similar indentation issue.

firmware/rhizome.ino
```diff
@@ -80,5 +90,13 @@ void loop() {
  * Delegation function that allows us to call the managed method for publishing periodic updates.
  */
 void doPeriodicUpdates() {
-    sproutList.publishPeriodicUpdates();
+	//Do not attempt to publish updates if disconnected from the cloud
+    if(Particle.connected() == false){
```
kyleoliveira: If `Particle.connected()` returns a boolean, you should just write this as `!Particle.connected()`

firmware/rhizome.ino
kyleoliveira: This is fine for now, but it would be great if we could store the `0x01` somewhere (as a const variable, naturally) so this is easier to understand. I don't know if we would put it in the menu object or some sort of simple settings object, but I could see some place for finding global constants like that being useful.

Just throwing that out there so we start thinking about it.

It may be as simple as having an Ohmbrewer_EEPROM class that extends EEPROM and defines hexes we use.

lib/Ohmbrewer_Menu.cpp
```diff
@@ -0,0 +1,156 @@
+#include "Ohmbrewer_Menu.h"
```
kyleoliveira, line 1: Since this file contains the methods for the Menu_WiFi class, it should be named "Ohmbrewer_Menu_WiFi.cpp" instead of "Ohmbrewer_Menu.cpp"

lib/Ohmbrewer_Menu.cpp
```diff
@@ -0,0 +1,156 @@
...
+ }
+ 
+/**
+ * Takes action when the select button is pressed in the Menu Screen
+ * The precise action will be determined by the individual menu screens
+ * @returns The time taken to run the method
+ */		
+int Ohmbrewer::Menu_WiFi::selectPressed(){
+	char status [40];
+	if(selectedOption == 0 && wiFiSetting == false) {  //Turn WiFi on
```
kyleoliveira: Just use `!wiFiSetting`

lib/Ohmbrewer_Menu.cpp
```diff
@@ -0,0 +1,156 @@
...
+				//Serial.println(wifiConnecting);
+				delay(100);
+			}
+			wifiConnecting = "Connected";
+			wifiConnecting.toCharArray(status, 40);
+			_screen->displayStatusUpdate(status);
+			delete wifiConnecting;
+		} 
+		
+	} else if(selectedOption == 1 && wiFiSetting == true) {
```
kyleoliveira: Just use `wiFiSetting`

lib/Ohmbrewer_Menu.h
```diff
@@ -0,0 +1,142 @@
...
+        std::vector<String> options;
+		
+		/**
+         * The currently selected option.
+         */
+		int selectedOption;
+	
+	};
+	
+	class Menu_WiFi : public Menu {
```
kyleoliveira: The Menu_WiFi class should get split out into its own .h file.

lib/Ohmbrewer_Screen.cpp
```diff
@@ -360,3 +369,87 @@ unsigned long Ohmbrewer::Screen::displayStatusUpdate(char *statusUpdate) {
...
+unsigned long Ohmbrewer::Screen::captureButtonPress() {
+    unsigned long start = micros();
+    char printx [10];
+	char printy [10];
+	char status [40];
+	
+	// a point object holds x y and z coordinates
+  	TSPoint p = ts->getPoint();
+  	//According to Particle forums, the read below is necessary to get touch to work on Photon
+  	ts->readTouchY();
```
kyleoliveira: Did you figure out if this read has to be done every single time you want a new point or if it can be done once after boot up? 'Cause if it's the latter, we should be able to move this into the constructor.

lib/Ohmbrewer_Screen.cpp
```diff
@@ -15,6 +16,14 @@ Ohmbrewer::Screen::Screen(uint8_t CS,
                           uint8_t RST,
                           std::deque< Ohmbrewer::Equipment* >* sprouts) : Adafruit_ILI9341(CS, RS, RST) {
     _sprouts = sprouts;
+	/** 
+    *  For better pressure precision, we need to know the resistance
+	*  between X+ and X- Use any multimeter to read it
+	*  Using value of 300 ohms across the X plate
```
kyleoliveira: Actually, this is using 285 ohms... Either fix the comment or fix the setting on line 24 :wink:

lib/Ohmbrewer_Screen.h
```diff
@@ -5,9 +5,18 @@
 #undef min
 #undef max
 #undef swap
+
+//Below definitions are for the touch screen
+#define XP A1
+#define YP A0
+#define XM A7
+#define YM A2
+
```
kyleoliveira, line 14: Change the `#define`'s for XP, YP, XM, and YM into static consts as we do for the other consts.

kyleoliveira:
Sorry it took so long to get around to reviewing this! Looks pretty good so far - most of these comments should be easy to rectify pretty quickly. I haven't tried building and booting this yet, but it seems reasonable on paper. I'll try to do so tonight just for completeness.

Post a comment when you've got a new commit and/or reply to anything that you're not sure about.

kyleoliveira:
Tested it out using a fork of [ScruffR/touch_4wire](https://github.com/ScruffR/touch_4wire) for the Touch_4Wire library, so if that's not the right one to use, let me know...

As you can see from the attached screenshot, the menu isn't rendering quite right (there's a bunch of ='s in the top right corner).

Additionally, pokes don't seem to be working. Judging by the fact that it keeps blinking that Select is being pressed, I think it may be a sensitivity thing. Once I hear back from @nischalvasant as to whether that's the right repo to use, I'll fiddle with it again and see if I can get it to pay just the right amount of attention. I know there are two things that can be modulated for this (the pressure ranges and the resistance setting in the constructor) - it seems like there should be some sort of reasonable ranges for these given that the screens are made by a reputable dealer.

Relatedly, when I do poke it, nothing is displayed... Maybe there's a case that's reversed somewhere?

firmware/rhizome.ino
chillinginchaos, line 84: I used tabs, but the indentation matches on my local editor... odd

firmware/rhizome.ino
chillinginchaos: Agreed... I planned on defining it in the Menu object, but extending EEPROM makes sense too..

firmware/rhizome.ino
chillinginchaos: Is this a coding standard? I dislike doing that because I see it as less clear.

lib/Ohmbrewer_Menu.cpp
chillinginchaos, line 1: This means we will end up with a massive amount of separate files for each Menu screen. What is the issue with having all of the Menu_Objects in the same .h and .cpp files?

lib/Ohmbrewer_Menu.cpp
chillinginchaos: Same question as before about coding standards.

lib/Ohmbrewer_Menu.cpp
chillinginchaos: Same question as before about coding standards.

lib/Ohmbrewer_Menu.h
chillinginchaos: Same question as before

lib/Ohmbrewer_Screen.cpp
chillinginchaos: Yes - it did not work when I placed it in the constructor. So far, I've only got it to work when it is at that spot.

lib/Ohmbrewer_Screen.h
chillinginchaos, line 14: It did not work for me when I did that...

chillinginchaos:
I'm not able to see an attachment, but are you using a photon or a core? If its a Core, I wonder if that is causing issues - but I'm not sure how to test that without purchasing one (is it even available?).

I'm 99% certain that's the repository I'm using, but I will check the origin from my folder on my local system and get back to you.

chillinginchaos:
It is definitely the ScruffR one.

  Fetch URL: https://github.com/ScruffR/touch_4wire.git
  Push  URL: https://github.com/ScruffR/touch_4wire.git

There is another one that I found but never used because I managed to get the other one to work 
  Fetch URL: https://github.com/wesner0019/Ported_Touch4Wire.git
  Push  URL: https://github.com/wesner0019/Ported_Touch4Wire.git

lib/Ohmbrewer_Menu.cpp
azyth, line 1: Wee can make a subdirectory to house all the menu files, and that will consolidate them and not clutter the main folder while at the same time allowing us to maintain the same conventions we have been using thus far.

lib/Ohmbrewer_Menu.h
chillinginchaos: Discussed with @azyth, will create a subfolder for menu classes to prevent clutter.

lib/Ohmbrewer_Menu.cpp
chillinginchaos, line 1: Discussed with @azyth, will create a subfolder for menu classes to prevent clutter.

kyleoliveira:
Ok, then I've got a fork of the ScruffR one set up for the team. I'll try out the new code sometime today.

Mostly looks good, though the indentation is still off. I don't know what you'll need to do in your IDE, but you can tell something is amiss when you look at it in the PR here.

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_WiFi.cpp
```diff
@@ -0,0 +1,167 @@
...
+				}
+				wifiConnecting.toCharArray(status, 40);
+				_screen->displayStatusUpdate(status);
+				//Serial.println(wifiConnecting);
+				delay(100);
+			}
+			wifiConnecting = "Connected";
+			wifiConnecting.toCharArray(status, 40);
+			_screen->displayStatusUpdate(status);
+			delete wifiConnecting;
```
kyleoliveira: You don't need to manually delete Particle Strings. They're objects and can handle their own dynamically allocated memory.

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_WiFi.cpp
```diff
@@ -0,0 +1,167 @@
...
+		wiFiSetting = false;
+		EEPROM.write(MENU_WIFI_ADDR, 0x00);
+		//disconnect from WiFi
+		Particle.disconnect();
+		wifiConnecting = "Disconnected.";
+		wifiConnecting.toCharArray(status, 40);
+		_screen->displayStatusUpdate(status);
+
+	}
+	_screen->reinitScreen();
```
kyleoliveira: You don't need to call reinitScreen() here because the screen gets refreshed on every loop. Calling this here causes the screen to blink (at least on my Photon).

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_WiFi.cpp
```diff
@@ -0,0 +1,167 @@
...
+	displayMenu();
```
kyleoliveira: You don't need to call displayMenu() here.

kyleoliveira:
If you can resolve the comments I added, I think you'll be good to go. I'm still getting some weird results with respect to button presses, but I think it's due to the touchscreen itself.

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_WiFi.cpp
chillinginchaos: When I don't call it here, I get weird (er) results. It kind of starts printing the menu again half way through the screen.

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_WiFi.cpp
chillinginchaos: Calling displayMenu() updates the current status of WiFi from "off" to "on" or vice versa.

chillinginchaos:
@kyleoliveira Is the formatting off when you download it? I'm not sure how to fix when it looks good on my machine. I'm using TextWrangler, an NP++ clone for mac.

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_WiFi.cpp
azyth: I was having this issue when I would try to display the screen again without allowing it to refresh, so you are trying to print twice in one cycle here correct?

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_WiFi.cpp
azyth: shouldn't all of the display stuff update on the next cycle anyways? or is this functionality only handled here?

kyleoliveira:
Yeah, formatting was not right when I downloaded it. Try opening it up in something other than TextWrangler (vim?) - I know I've seen this sort of thing happen with other folks using that program. If you want to get it sorted in TextWrangler, you'll want to find settings in the program for both how indents are displayed and how they're added when you hit tab. In both cases, set it to 4 for this project. Then you should (hopefully) be able to see the difference. Otherwise, let's set up a time to sort it out.

kyleoliveira:
See #74 for more info, but once @nischalvasant is ready, this should be good for merge.

PR #74, opened 2016-03-17
Adds a RuntimeSettings class for keeping track of global settings and refactors the menu classes to remove settings tracking concerns and generally be more stateful.

kyleoliveira:
@nischalvasant This is basically what I was talking about yesterday. It did end up being a bit much, but I think that's because the original task should have probably been split up a bit (i.e. a PR for just poking something, a PR for saving settings to EEPROM, etc.)

Unfortunately, even though this compiles, I can't seem to get it to work on my Rhizomes. The touchscreen always seems to be pulling a coordinate at the far corner of the screen. Between this and the varying calibration for the screen pressure is making me wonder if we should have gone with hardware buttons for the input...

Could you try this out on your end and see if you're getting the same results? Either way, take a look and let me know what you think and/or if you know the what's causing the problem.

@azyth Look too :grinning:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,8 @@ set(SOURCE_FILES
     ../lib/Ohmbrewer_Heating_Element.cpp
     ../lib/Ohmbrewer_Menu.h
     ../lib/Ohmbrewer_Menu.cpp
+    ../lib/Ohmbrewer_Menu/Ohmbrewer_Menu_Home.h
+    ../lib/Ohmbrewer_Menu/Ohmbrewer_Menu_Home.cpp
     ../lib/Ohmbrewer_Menu/Ohmbrewer_Menu_WiFi.h    
     ../lib/Ohmbrewer_Menu/Ohmbrewer_Menu_WiFi.cpp
     ../lib/Ohmbrewer_Onewire.cpp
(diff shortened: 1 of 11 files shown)


chillinginchaos:
Looks like a good way to do it. I only made a weak attempt at  building out the persistence bit anyway. I'll try compiling it tonight. The reading being stuck at one corner of the screen issue is the one that was fixed by having the readY in the particular location. You can also try to use the "alternative" repository I had linked. I have not tried that so far.

kyleoliveira:
Aaaah, ok. I'll take a look at adding that readY thingy back in when I get off work and see if that helps. If I haven't already pushed that up by the time you get a chance to take a look, add it in locally and let me know if that was the problem.

kyleoliveira:
Actually the readTouchY call was in there. I think the issue is the library - swapping the alternative on in seems to help. I'm going to debug for another half hour or so and then I'm going to need to work on other stuff. If you don't have a chance to get it working today, I'll pick up again tomorrow evening.

kyleoliveira:
Worked it out with @nischalvasant over Hangouts. With these settings, this seems to work on his Rhizome. Some calibration will be needed from Rhizome to Rhizome, specifically modifying the following settings:
Ohmbrewer_Screen.h
- MINPRESSURE
- MAXPRESSURE
  Ohmbrewer_Screen.cpp
- In the Screen constructor, the last parameter to the Touchscreen constructor (resistance across the screen in ohms)

Notably, it seems that my white Rhizome's resistance is reversed compared to these settings. We'll want to add additional issues for calibration (or otherwise go to hardware buttons).

PR #75, opened 2016-03-29
Small fix to make the screen pressure range make more sense.

kyleoliveira:
@nischalvasant Please confirm that this still works for your Rhizome. I don't know why I swapped that around.

### DIFF
--- a/lib/Ohmbrewer_Screen.cpp
+++ b/lib/Ohmbrewer_Screen.cpp
@@ -407,13 +407,13 @@ unsigned long Ohmbrewer::Screen::captureButtonPress() {
   
     // we have some minimum pressure we consider 'valid'
     // pressure of 0 means no pressing!
-    if (p.z > MINPRESSURE || p.z < MAXPRESSURE) {
+    if (p.z < MINPRESSURE || p.z > MAXPRESSURE) {
         displayStatusUpdate("                                        ");
         return micros() - start;
     }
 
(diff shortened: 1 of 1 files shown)


lib/Ohmbrewer_Screen.cpp
```diff
@@ -456,9 +456,9 @@ unsigned long Ohmbrewer::Screen::captureButtonPress() {
         displayStatusUpdate(status);
     }
     
-//    Serial.print("X = "); Serial.print(p.x);
-//    Serial.print("\tY = "); Serial.print(p.y);
-//    Serial.print("\tPressure = "); Serial.println(p.z);
+    Serial.print("X = "); Serial.print(p.x);
```
kyleoliveira, line 459: I can comment these out again, but I was thinking that since we're doing a lot of touchscreen work we might just leave them as-is for now.

chillinginchaos:
Compiles and works OK on mine. Although there is also that "-35" hack from the original code that I needed to use to get the touches to be accurate. Without it, the touches seem to recognize slightly to the right of the actual touch.

kyleoliveira:
I've added the -35 back in. Let me know if we're good to go and I'll merge it. :smile:

chillinginchaos:
Looks good, go ahead!

kyleoliveira:
:zap: :zap: :zap:

Issue #32, opened 2015-10-31
Make ID's transparent to the User


kyleoliveira:
@Ohmbrewer/web-team 
So far we've been manually setting the IDs and we've implemented a few solutions for subsystems in Thermostat and RIMS. However, it would make it a lot easier for the Ohmbrewer app to talk to the Rhizome if we made it more transparent. Specifically, I would like to remove the Equipment ID from all constructors and set it according to a unique property of the Equipment in question. For most of the Equipment, this would be a pin number. For the Onewire sensors, we'd have to get a little creative. I'm thinking:
- Relay, Pump & Heating Element: power pin
- Thermostat: Element power pin
- RIMS: Thermostat's ID
- Temperature Sensor: Bus index?

Thoughts? I feel like this will make communications much easier to automate.

azyth:
I think this is a good idea in general but there will be some snags. 

all relays are inherently flexible as to which pin they use. so that might get iffy

in theory if one was unused you could just use the other since it would have to be unique since we only have one of each pin, but that is less specific and might get confusing.

kyleoliveira:
From a programming perspective, though aren't we requiring that, at least, the "power" pin be set for all Relay-based Equipment? I thought we're using the "control" pin for a particular purpose (modulation) and it wouldn't be appropriate to swap them. 
I mean, it might not matter physically, but we do want to make the distinction in what the Rhizome will be doing with a particular pin, right?

azyth:
nope, relays have the control pin set by default, but the power pin may or may not be set, and pumps (currently) only set the power pin.

kyleoliveira:
We talked it over via Hangouts and it seems like these are the pins that we could require always be used:
- Relay & Pump: Power Pin 
- Heating Element: Control Pin
- Thermostat: Element ID
- RIMS: Thermostat's ID

Still not 100% on the Temperature Sensor, though. I'd like to try Bus index and see how that goes because it seems like it would be fairly straight-forward to get the Probe info screen to display in that order (if it isn't already) and then the user wouldn't have to keep track of the OneWire address.

Sound good?

azyth:
bus index changes every time it queries. so I think the best way would be to keep track internally of the HEx addresses and somehow just convert them to an equivalent int or something for display to the user. This will involve some creative code i think.

kyleoliveira:
Damn, I was hoping that as long as you didn't change the wiring they'd always return in the same order.

[...]

PR #76, opened 2016-03-29
Refactors Relay to use always use Control Pin instead of always using Power Pin.

### LINKED ISSUE #70: Refactor Pump/Relay to use Control Pin instead of Power Pin
As discussed in #32 , Pump and Relay should use Control Pin as their one pin instead of Power pin.

kyleoliveira:
Resolves #70.

@azyth Take a look and let me know if I broke something inadvertently or didn't do enough. See the linked Issue for why I'm doing this.

### DIFF
--- a/lib/Ohmbrewer_Relay.cpp
+++ b/lib/Ohmbrewer_Relay.cpp
@@ -6,29 +6,28 @@
 /**
  * Constructor
  * @param id The Sprout ID to use for this piece of Equipment
- * @param powerPin - Single speed pump will only have PowerPin - on/off line. Digital pin number X.
+ * @param controlPin - Single speed pump will only have ControlPin - on/off line. Digital pin number X.
 
  */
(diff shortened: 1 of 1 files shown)


azyth:
Just so I am clear, this is the issue we created last semester and just never got around to it?  

What you have done seems fine.  I need to think about it a little though for any other ramifications

kyleoliveira:
Yeah, pretty much. It goes back to #32. If we always have a particular pin set for the equipment, we can then use that pin number as the equipment ID. In part of that discussion, we determined that pretty much everything other than OneWire stuff should always have a Control Pin.

As for ramifications, we'll need to verify on the Ohmbrewer side that it's still sending messages as expected. Since Pump was only using one pin and relying on Relay's one pin constructor, I think this handles any side-effects on the Rhizome side.

kyleoliveira:
@azyth Let me know if this is good to go, it's holding up another, more important, PR.

kyleoliveira:
Seeing no complaints since this has been up for a while, I'm merging it.

PR #77, opened 2016-04-08
Refactors out manual IDs, replacing them with implementations of getID.

### LINKED ISSUE #32: Make ID's transparent to the User

kyleoliveira:
Resolves #32.

Note: this changes the /add Particle function's API (no longer need to pass in the ID).

Waiting on #76 before this can be merged in. We'll need to update the appropriate communication methods in Ohmbrewer to accommodate the API change - I'll make an issue for that before this gets merged in.

### DIFF
--- a/lib/Ohmbrewer_Equipment.cpp
+++ b/lib/Ohmbrewer_Equipment.cpp
@@ -5,7 +5,6 @@
  * Constructor
  */
 Ohmbrewer::Equipment::Equipment() {
-    _id = 0;
     _stopTime = 0;
     _state = false;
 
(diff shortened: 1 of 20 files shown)


kyleoliveira:
Haven't tested this yet, but it does compile. @azyth take a look at this and let me know if there are any issues.

kyleoliveira:
Ok, tested it out and made some small tweaks. Most importantly, I removed the block that was stopping the periodic timer because it caused a hard fault when we tried subsequently reset it in the global periodic update function. Since there's not way to test whether the timer has been started or stopped, we can't just put an if-statement around the reset so this seemed like the easiest solution.

That said, this seems to work for me. Because the buttons don't work for me, I had to comment out:
- rhizome.ino line 62
- Ohmbrewer_Screen.cpp lines 131-133,159

I think the latter is actually a bug (we're starting on the Home menu screen, so everything should display by default), so I'll add an issue for it.

All of that said, this works for me.

kyleoliveira:
@azyth & @nischalvasant Please take a look at this if you haven't already. It's been up for a bit, so I'm going to merge it in before the end of the weekend if I don't hear anything.

azyth:
This looks OK to me. If it is working on the unit itself then I think we are OK

kyleoliveira:
I added a new issue to the Ohmbrewer issue queue so we don't forget to update that...

Issue #79, opened 2016-04-12
Need a way to select temperature units


chillinginchaos:
Switch between displaying C and F to the user.

PR #80, opened 2016-04-12
Adds Main Menu and allows selection between WiFi menu and Temperature Unit Selection

### LINKED ISSUE #78: Add Main Menu
Create a main menu as the base of a menu heirarchy

chillinginchaos:
Resolves #78 #79

### DIFF
--- a/firmware/rhizome.ino
+++ b/firmware/rhizome.ino
@@ -64,6 +64,12 @@ void setup() {
 	//initialize the Dallas Onewire bus pin - Digital 0
 	ow_setPin(D0); //This should later be accomplished by equipment setup OR constructor
 
+    if(!Particle.connected()){
+        if(EEPROM.read(Ohmbrewer::RuntimeSettings::WIFI_STATUS_ADDR) == Ohmbrewer::RuntimeSettings::EEPROM_WIFI_STATUS_ON){
+            //WiFi is not connected and should be - attempt to connect
+            Particle.connect();
+        }
+    }
 	// Turn on the display
 	screen.initScreen();
 	
@@ -95,8 +101,7 @@ void loop() {
 void doPeriodicUpdates() {
 	//Do not attempt to publish updates if disconnected from the cloud
     if(!Particle.connected()){
-        // TODO: Define WiFi Address location properly
-        if(EEPROM.read(1) == Ohmbrewer::RuntimeSettings::EEPROM_WIFI_STATUS_OFF){
+        if(EEPROM.read(Ohmbrewer::RuntimeSettings::WIFI_STATUS_ADDR) == Ohmbrewer::RuntimeSettings::EEPROM_WIFI_STATUS_ON){
             //WiFi is not connected and should be - attempt to connect
             Particle.connect();
         }
--- a/lib/Ohmbrewer_Menu/Ohmbrewer_Menu_Home.cpp
+++ b/lib/Ohmbrewer_Menu/Ohmbrewer_Menu_Home.cpp
@@ -9,7 +9,11 @@
  * @param settings A pointer to the RuntimeSettings object.
  */
 Ohmbrewer::MenuHome::MenuHome(Ohmbrewer::Screen *screen, Ohmbrewer::RuntimeSettings *settings) : Ohmbrewer::Menu::Menu(screen, settings) {
-    // Nothing new to do
+    options.push_back("Temp Sensors");
+    options.push_back("RIMS");
+    options.push_back("Thermostats");
+    options.push_back("Pumps");
+    _selectedOption = 0;
 }
 
 /**
@@ -17,7 +21,25 @@ Ohmbrewer::MenuHome::MenuHome(Ohmbrewer::Screen *screen, Ohmbrewer::RuntimeSetti
  * @todo Refactor the display drawing stuff that's currently in Screen into this
  */
 void Ohmbrewer::MenuHome::displayMenu() {
-    // Do nothing for now...
+
+    if(_selectedOption == 0) {
+        _screen->displayTemps();
+    } else if (_selectedOption == 1){
+        _screen->displayRIMS();
+    } else if (_selectedOption == 2) {
+        _screen->displayThermostats();
+    } else if (_selectedOption == 3) {
+        _screen->displayPumps();
+    }
+
+
+
+
+    //    // These are the old, single type displays, for easy debugging.
+    //_screen->displayManualRelays();
+    //    displayRelays();
+    //    displayHeatingElements();
+    //    displayPumps();
 }
 
 /**
@@ -25,5 +47,9 @@ void Ohmbrewer::MenuHome::displayMenu() {
  * The precise action will be determined by the individual menu screens
  */
 void Ohmbrewer::MenuHome::menuPressed() {
-    _screen->setCurrentMenu(_children[WIFI_MENU]);
+    _screen->setCurrentMenu(_children[MAIN_MENU]);
+    _selectedOption = 0;
+    _screen->reinitScreen();
 }
+
+
--- a/lib/Ohmbrewer_Menu/Ohmbrewer_Menu_TempUnit.cpp
+++ b/lib/Ohmbrewer_Menu/Ohmbrewer_Menu_TempUnit.cpp
@@ -1,0 +1,62 @@
+#include "Ohmbrewer_Menu_TempUnit.h"
+#include "Ohmbrewer_Runtime_Settings.h"
+#include "Ohmbrewer_Screen.h"
+
+/**
+ * Constructors
+ * @param screen A pointer to the Screen object.
+ * @param settings A pointer to the RuntimeSettings object.
+ */
+Ohmbrewer::MenuTempUnit::MenuTempUnit(Ohmbrewer::Screen *screen, Ohmbrewer::RuntimeSettings *settings) : Ohmbrewer::Menu::Menu(screen, settings) {
+    options.push_back("Celsius");
+    options.push_back("Fahrenheit");
+    _selectedOption = 0;
+
+}
+
+/**
+ * Displays the Temp Unit Menu and options
+ */
+void Ohmbrewer::MenuTempUnit::displayMenu() {
+
+    _screen->resetTextSize();
+    _screen->resetTextColor();
+
+    // Print the section title
+    _screen->print("======= MENU =======");
+    _screen->printMargin(2);
+    _screen->print(" ---- Temp Unit --- ");
+    _screen->printMargin(2);
+    //Print current status
+    _screen->print("Selected Unit: ");
+    _screen->writeDegree();
+    //_screen->printMargin(2);
+    if(_settings->isTempUnitCelsius()){
+        _screen->println("C");
+    } else {
+        _screen->println("F");
+    }
+
+    printOptions();
+
+}
+
+/**
+ * Takes action when the select button is pressed in the Menu Screen
+ * The precise action will be determined by the individual menu screens
+ * @returns The time taken to run the method
+ */
+void Ohmbrewer::MenuTempUnit::selectPressed() {
+
+    if(_selectedOption == 0) {
+        //Degrees Centigrade!
+        _settings->setTempUnitAndSave(UNIT_CELSIUS);
+
+
+    } else if(_selectedOption == 1) {
+        // Degrees Fahrenheit
+        _settings->setTempUnitAndSave(UNIT_FAHRENHEIT);
+    }
+}
+
+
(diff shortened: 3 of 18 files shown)


lib/Ohmbrewer_Menu/Ohmbrewer_Menu_TempUnit.cpp
```diff
@@ -0,0 +1,69 @@
...
+        _settings->setTempUnitAndSave(UNIT_FAHRENHEIT);
+    }
+}
+
+/**
+ * Takes action when the menu button is pressed in the Menu Screen
+ * The precise action will be determined by the individual menu screens
+ */
+void Ohmbrewer::MenuTempUnit::menuPressed() {
+    _screen->setCurrentMenu(_parent);
```
kyleoliveira: It almost seems like this should be the default action and just have the HomeMenu override it with an empty function. Does that seem reasonable?

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_TempUnit.h
```diff
@@ -0,0 +1,61 @@
...
+#include <vector>
+
+
+namespace Ohmbrewer {
+
+    class MenuTempUnit : public Menu {
+
+    public:
+
+        static const bool UNIT_CELCIUS = true;
```
kyleoliveira: Should be 'UNIT_CELSIUS'.

Also, since this is namespaced, you could just go with 'CELSIUS'.

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_TempUnit.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "Ohmbrewer_Runtime_Settings.h"
+#include "Ohmbrewer_Screen.h"
+
+/**
+ * Constructors
+ * @param screen A pointer to the Screen object.
+ * @param settings A pointer to the RuntimeSettings object.
+ */
+Ohmbrewer::MenuTempUnit::MenuTempUnit(Ohmbrewer::Screen *screen, Ohmbrewer::RuntimeSettings *settings) : Ohmbrewer::Menu::Menu(screen, settings) {
+    options.push_back("Celcius");
```
kyleoliveira: "Celsius"

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_WiFi.cpp
```diff
@@ -12,7 +12,7 @@
     //Set up list of options in vector
     options.push_back("WiFi On");
     options.push_back("WiFi Off");
-    _selectedOption = _settings->isWifiOn();
+    _selectedOption = 0;
```
kyleoliveira, line 15: Was this just a debugging change?

lib/Ohmbrewer_Runtime_Settings.h
```diff
@@ -69,13 +74,32 @@ namespace Ohmbrewer {
...
 
         /**
          * Whether the WiFi should be on (true) or off (false)
          */
         bool _wifiStatus;
 
+        /**
+         * Whether the Temp Unit should be Celcius (true) or Fahrenheit (false)
+         */
+        bool _celcius;
```
kyleoliveira: _celsius

Instead of commenting on all of these, from here on out let's just assume you'll find and replace throughout the project :wink:

lib/Ohmbrewer_Screen.cpp
```diff
@@ -31,7 +33,11 @@ Ohmbrewer::Screen::Screen(uint8_t CS,
 
     // Build the menu tree
     _homeMenu = new MenuHome(this, settings);
-    _homeMenu->addChild(new MenuWiFi(this, settings));
+    _mainMenu = new MenuMain(this, settings);
```
kyleoliveira: In the next batch of menus you might want to refactor out this tree assignment into its own function that gets called from the constructor.

lib/Ohmbrewer_Screen.cpp
```diff
@@ -398,22 +382,21 @@ unsigned long Ohmbrewer::Screen::captureButtonPress() {
...
     // a point object holds x y and z coordinates
     TSPoint p =_ts->getPoint();
     //According to Particle forums, the read below is necessary to get touch to work on Photon
     _ts->readTouchY();
 
-  
     // we have some minimum pressure we consider 'valid'
     // pressure of 0 means no pressing!
-    if (p.z > MINPRESSURE || p.z < MAXPRESSURE) {
+    if (p.z < MINPRESSURE || p.z > MAXPRESSURE) {
```
kyleoliveira, line 390: If this is the same as #75, maybe we can get that confirmed and merged in? :wink:

lib/Ohmbrewer_Screen.h
```diff
@@ -213,6 +213,11 @@ namespace Ohmbrewer {
              */
             Menu* _homeMenu;
 
+           /**
```
kyleoliveira: Actually, we shouldn't need to have an instance variable for the main menu. We only need to keep track of the root (HomeMenu) and the current menu. I think that's already implemented...

lib/Ohmbrewer_Temperature_Sensor.h
```diff
@@ -178,6 +184,11 @@ namespace Ohmbrewer {
              */
             Probe* _probe;
 
+            /**
```
kyleoliveira: Rather than duplicating the value of the display setting from the singleton settings object in every TemperatureSensor instance, write a Temperature#toStr(bool units) method that calls either Temperature#toStrC() or Temperature#toStrF() based on the units parameter (defaulting to celsius).

That should save on the number of writes to memory vs. what you've got going on now and avoid the penalty for setting that on every TemperatureSensor every time we refresh the display.

You may need to add a getSettings() method to Screen to accomplish this... (i.e. have the TemperatureSensor grab _screen->getSettings()->isTempCelsius() or some such)

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_TempUnit.cpp
chillinginchaos: For menu pressed? Agreed. I meant to implement that at some point, just forgot.

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_WiFi.cpp
chillinginchaos, line 15: No, I wanted it to start at the first option every time. I think thats how I would expect a menu to work.

lib/Ohmbrewer_Screen.cpp
chillinginchaos, line 390: Commented on it :+1:

lib/Ohmbrewer_Screen.h
chillinginchaos: How to add multiple children when each is manually being added though?

lib/Ohmbrewer_Temperature_Sensor.h
chillinginchaos: Sure, thats fine, but I don't see how we are writing to EEPROM any differently. I am only reading here, and that too from the boolean, so there should be no real difference between the method you are suggesting other than it being "cleaner"

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_TempUnit.cpp
kyleoliveira: Yup yup. Go for it.

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_WiFi.cpp
kyleoliveira, line 15: What if when the wifi what shut off when the Rhizome was last running? We want to load the settings from EEPROM, right? The constructor for _settings will load the setting, but we need to pass it into the menu to get the view to show what's actually selected.

This would only really be an issue on the very first time the user goes to this menu. The change would make it so in this situation the wifi would actually be off, but the state of the menu wouldn't change until the user selects an option.

lib/Ohmbrewer_Screen.cpp
kyleoliveira, line 390: Kickass

lib/Ohmbrewer_Screen.h
kyleoliveira: The same way you're doing it, except mainMenu would be a local variable instead of an instance variable. It should work because it's a pointer.

This actually highlights a bug though! We don't have a destructor for this class (we just rely on the parent class's destructor), which means that if somehow the screen was deleted, the touchscreen and menu instance variables would turn into a memory leak (the other variables are managed separately). We've never noticed before because we never delete the screen. We should still provide that though. You can either take a whack at it here, or open an issue that references this since it's kind of an increase in scope.

lib/Ohmbrewer_Temperature_Sensor.h
kyleoliveira: Correct, it doesn't affect EEPROM operation. I meant "memory" in terms of RAM at runtime. Over in Screen::displayTemps() you're setting _displayCelsius in every TemperatureSensor's Temperature object on every screen refresh.

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_WiFi.cpp
chillinginchaos, line 15: This bit of code later achieves this:

```
if(_settings->isWifiOn()){
    _screen->println("ON ");
} else {
    _screen->println("OFF");
}
```

lib/Ohmbrewer_Screen.h
chillinginchaos: Changed the mainMenu to be a local variable.

Created an issue for the destructor bit

firmware/rhizome.ino
```diff
@@ -21,6 +21,7 @@
 #include "Ohmbrewer_Runtime_Settings.h"
 //external libraries
 #include "onewire.h"
+#include "Ohmbrewer_Runtime_Settings.h"
```
kyleoliveira: Extra include...

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_Home.cpp
```diff
@@ -9,21 +9,44 @@
  * @param settings A pointer to the RuntimeSettings object.
  */
 Ohmbrewer::MenuHome::MenuHome(Ohmbrewer::Screen *screen, Ohmbrewer::RuntimeSettings *settings) : Ohmbrewer::Menu::Menu(screen, settings) {
-    // Nothing new to do
+    options.push_back("Temp Sensors");
```
kyleoliveira, line 12: If we're switching things up to work this way, we should include an option for standalone pumps.

lib/Ohmbrewer_Screen.cpp
```diff
@@ -128,44 +139,12 @@ unsigned long Ohmbrewer::Screen::refreshDisplay() {
 
     displayHeader();
 
-    if(!_currentMenu->isHome()) {
+    //if(!_currentMenu->isHome()) {
```
kyleoliveira: You should be able to clear out the commented stuff if this works.

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_WiFi.cpp
kyleoliveira, line 15: :+1:

kyleoliveira:
I think you just need to clean up some spacing and commented out stuff, and add a thingy to display standalone pumps (in the menu you're replacing the home screen with) and this will be ready. I haven't tested it out or compiled, but it looks about right.

firmware/rhizome.ino
chillinginchaos: LOL. I haven't touched the rhizome.ino file at all this iteration.

lib/Ohmbrewer_Menu/Ohmbrewer_Menu_Home.cpp
chillinginchaos, line 12: Done.

lib/Ohmbrewer_Screen.cpp
chillinginchaos: Done.

chillinginchaos:
New changes pushed.

kyleoliveira:
:zap: :metal: :zap:

Squash into one commit and you'll be go for merge.

chillinginchaos:
Resolves #78 #79

PR #84, opened 2016-04-28
Adds index variable exposed via particle cloud REST functionality

chillinginchaos:
This adds an index variable with a list of equipment, their id, type and current status. Will add pins in a future PR.

https://docs.particle.io/reference/firmware/photon/#particle-variable-

### DIFF
--- a/lib/Ohmbrewer_Rhizome.h
+++ b/lib/Ohmbrewer_Rhizome.h
@@ -147,6 +147,12 @@ namespace Ohmbrewer {
          */
         Timer* _periodicUpdateTimer;
 
+        /**
+         * Keeps a json format index of all registered equipment for easy access
+         * to any requesting application. Exposed via particle.variable
+         */
+        String _index;
+
 
     private:
 
@@ -247,6 +253,17 @@ namespace Ohmbrewer {
          * @return Error or success code, according to the requirements specified by addSprout
          */
         int addRIMS(char* params);
+
+        /**
+         * Adds a sprout to the list and rebuids the index
+         * @param equipment The sprout being added
+         */
+        void saveNewSprout(Equipment* sprout);
+
+        /**
+         * Rebuilds index based on current list of equipment
+         */
+        void rebuildIndex();
     };
 };
 
--- a/lib/Ohmbrewer_Rhizome.cpp
+++ b/lib/Ohmbrewer_Rhizome.cpp
@@ -24,6 +24,8 @@ Ohmbrewer::Rhizome::Rhizome(Timer *put) {
     Particle.function("add", &Rhizome::addSprout, this);
     Particle.function("update", &Rhizome::updateSprout, this);
     Particle.function("remove", &Rhizome::removeSprout, this);
+    Particle.variable("index",_index);
+
 }
 
 /**
@@ -198,6 +200,7 @@ int Ohmbrewer::Rhizome::removeSprout(String argsStr) {
         if (((*itr)->getID() == id) && (type.equalsIgnoreCase((*itr)->getType()))) {
             _sprouts->erase(itr);
             delete params;
+            rebuildIndex();
             _screen->initScreen(); // Gotta do this to clear artifacts from the screen
             return id; // Success!
         }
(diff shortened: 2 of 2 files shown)


kyleoliveira:
Looks pretty good, but I don't see anything that actually triggers rebuildIndex()... or am I missing something?

chillinginchaos:
"I'm just going to copy paste since the changes are super localized"

...famous last words :'(

I'll fix this tonight. Basically in every add function, there is a rebuildIndex after successful add.

kyleoliveira:
Haha yeah, classic... I'll check again later then.

kyleoliveira:
You also need to rebuild the index when sprouts are removed.

What do you think about adding two more methods?:

``` c++
int Ohmbrewer::Rhizome::addSprout(Equipment* sprout) {
  _sprouts->push_back(sprout);
  rebuildIndex();
}

int Ohmbrewer::Rhizome::removeSprout(int sproutId) {

    std::deque<Ohmbrewer::Equipment*>::iterator itr = _sprouts->begin();
    for (itr; itr != _sprouts->end(); itr++) {
        if (((*itr)->getID() == sproutId) && (type.equalsIgnoreCase((*itr)->getType()))) {
            _sprouts->erase(itr);
            // Note that we're not passing the params variable in, so that memory will still need to be managed in the method that calls this...
            rebuildIndex(); // Rebuild the index
            _screen->initScreen(); // Gotta do this to clear artifacts from the screen
            return id; // Success!
        }
    }
  return -1;
}

```

No guarantees those would work as-is since I'm doing this off the top of my head, naturally, but you get the idea. Those would cut down on the number of times the `...->push_back...rebuildIndex...` lines are used and give a clear alternative for managing sprouts when you're not dealing with with strings.

lib/Ohmbrewer_Rhizome.h
```diff
@@ -247,6 +253,17 @@ namespace Ohmbrewer {
          * @return Error or success code, according to the requirements specified by addSprout
          */
         int addRIMS(char* params);
+
+        /**
+         * Adds a sprout to the list and rebuids the index
+         * @param equipment The sprout being added
+         */
+        void saveNewSprout(Equipment* sprout);
```
kyleoliveira, line 261: Let's call this `addSprout`.

chillinginchaos, line 261: There is already a function named addSprout. The compiler wasn't happy when I did that ;)

kyleoliveira, line 261: Weird... Are you sure the name was the problem? It _should_ work, given that one takes a Equipment\* and one takes a String: http://www.cplusplus.com/doc/tutorial/functions2/

chillinginchaos, line 261: Yeah it was the bit where we pass a reference to the function to the Particle.function call.

kyleoliveira, line 261: Ah, ok. So it looks like we probably just need to be more specific in the Particle call: http://stackoverflow.com/questions/17874489/disambiguate-overloaded-member-function-pointer-being-passed-as-template-paramet

chillinginchaos, line 261: I'm not sure I understand the logic behind naming it the same though. Why is saveNewSprout not adequate?

kyleoliveira, line 261: Mostly to minimize synonyms. This method does the same thing as the existing addSprout method, except the existing one includes the extra handling necessary for Strings. Both methods just provide the necessary machinations to put a particular type of object in the Sprouts list. As I see it, this is just the sort of use case that [method overloading](http://www.tutorialspoint.com/cplusplus/cpp_overloading.htm) was designed for.

kyleoliveira, line 261: Based on our conversation earlier this week (and that it can really go either way and the semester is almost over), let's just go with the name as you have it if you haven't already made the change. Renaming later isn't hard if it really matters.

I think you're ready to squash and merge this.

chillinginchaos, line 261: Just saw this now - squashing and merging.

kyleoliveira, line 261: :+1:

PR #85, opened 2016-05-03
Adds classes for the various error codes shown when interacting with the REST API.

kyleoliveira:
Should help for readability.

@Ohmbrewer/hops-team Whatcha think? Another low priority one, but I think it does make things easier to read and centralizes the location for setting the actual values used when we throw an error code.


azyth:
I like it, cleans it up nicely

PR #86, opened 2016-05-26
Adds the ability to remove all Sprouts or all of a given type by specifying all,all or TYPE,all when calling /remove

### LINKED ISSUE #60: Add "all" as an option for the /remove Particle.function
It would be very handy to be able to send in `all` instead of a specific ID to /remove and have it remove all Sprouts of a given type.

If we could send in `all,all` and clear all Sprouts, that would be even handier.

kyleoliveira:
Resolves #60 .

It builds, but I haven't tested it because I didn't tote my Rhizome to Ithaca with me. If someone could give it a go, I'd appreciate it. 

To test, add two or more pumps, two or more temperature sensors, and one other piece of equipment (or something along those lines). Then try:
`particle call RHIZOME remove temp,all`
and verify that all temperature sensors were removed. Then
`particle call RHIZOME remove all,all`
and verify that all remaining equipment was removed. It may also be interesting to try running both versions of the call when there is no equipment attached to verify that it returns the code that the documentation advertises.

@Ohmbrewer/hops-team Whatcha think?

### DIFF
--- a/lib/Ohmbrewer_Rhizome.cpp
+++ b/lib/Ohmbrewer_Rhizome.cpp
@@ -23,8 +23,8 @@ Ohmbrewer::Rhizome::Rhizome(Timer *put) {
 
     Particle.function("add", &Rhizome::addSprout, this);
     Particle.function("update", &Rhizome::updateSprout, this);
-    Particle.function("remove", &Rhizome::removeSprout, this);
-    Particle.variable("index",_index);
+    Particle.function("remove", &Rhizome::removeSprouts, this);
+    Particle.variable("index", _index);
 
 }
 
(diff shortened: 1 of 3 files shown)


kyleoliveira:
Actually, looks like I reversed the parameters for the type case. Will update a little later....

kyleoliveira:
Alrighty, fixed up.

kyleoliveira:
Rebased it off of master and tested it out. It should be working now. I'll wait ~24 hours and merge it if I don't see any complaints.

kyleoliveira:
Tidied up Rhizome::removeAllSprouts(String). Tested and still works right.

PR #87, opened 2016-05-27
Adds a constant to represent invalid temperature readings.

kyleoliveira:
@Ohmbrewer/hops-team 
This is one of those minor little tweaks that should help with consistency. I've added a constant to represent invalid temperature values (set to the existing implementation of "-69.0") and added a default constructor for the Temperature object that doesn't take any parameters.


azyth:
looks good to me,

PR #90, opened 2016-07-06
Adds Doxygen support.

### LINKED ISSUE #41: Build Doxygen documentation
Seeing as how we've been pretty good about our inline documentation, before v1.0 we should see if we can run Doxygen on our Ohmbrewer classes and extract supporting documentation directory.

kyleoliveira:
Resolves #41 

Documentation is generated by running `doxygen` from the project root and produces HTML output in the `/doc` directory.

The documentation generated currently says we're on v1.0, but we can change that when we decide what version we're actually at.

Also adds a .keep file in `/build` to force that directory to exist.


kyleoliveira:
See http://www.stack.nl/~dimitri/doxygen/manual/starting.html for information on installing `doxygen`, btw.

azyth:
cool, looks good.

PR #91, opened 2017-07-18
New version for updated rhizome with non touch screen

azyth:
ok, so this i broken right now..... a bunch of compile errors dealing with dependiecies and functions apparently no longer declared. 

you do have to add the actual package files to the parent directory as well. 

@kyleoliveira  I dont even remeber what the function call to begin() (line 68) was supposed to do, and was that something built in to another package or was that our function. but that appears to be a problem chiild in here. (one of many probably)


kyleoliveira:
Seems pretty reasonable so far. Definitely don't want to merge it until it's working though 😉 
If you want, you can always have a side branch off of master that you could build into incrementally until everything is ready to officially replace the 1st gen screen.

